Issue #2, opened 2023-01-04
TODO's


v0l:
Here are a list of bugs and features (& feature improvements) which need to be implemented (in no order):

## New Features
- [ ] DM's
- [x] Delete notes
- [ ] Mentions (Write mentions, with auto-complete)
- [x] Relay list management (Separate page from tab on profile)
- [x] NIP-05 Validation
- [x] Global tab (Currently exists but only for logged out users)
- [ ] Show newest notes (Don't auto load new notes onto the page, hide them behind button)
- [ ] Profile: 
  - [ ] Reaction Tab
  - [x] Followers Tab
  - [x] Follows Tab
- [ ] Notifications (Partially implemented, need service-worker)
- [ ] Link previews (OpenGraph)
- [ ] NIP-05 Shop (Allow creation of NIP-05 links directly from profile page (free/paid services to be connected))
- [ ] User settings:
  - [ ] Allow users to selectively disable reactions [Bandwidth] 
  - [ ] Allow users to selectively disable automatically loading media [Bandwidth/Privacy]
- [ ] Relay load balancing (Epic)
  - [ ] Load balance subscriptions by "splitting" filters where `ids` or `authors` are specified
- [ ] Hash tags:
  - [x] Highlight hashtags
  - [ ] Make hashtags searchable
- [ ] Markdown parsing:
  - [ ] Heading / text formatting / code blocks

## Feature Improvements
- [x] Thread improvements:
  - [x] Group replies into tree structure
  - [x] Visually show thread trees
- [ ] New user UX:
  - [ ] Export keys and explain key security
- [x] Reactions:
  - [x] Show already reacted notes with red icon
- [ ] Timeline improvements:
  - [ ] Show reactions in timeline views
  - [ ] Parse mentioned `pub_key` metadata from timeline (Currently showing names if already cached)
- [ ] Notes:
  - [x] Show time as `HH:mm` when > 60 mins ago
  - [x] Show time as `YYYY-MM-dd HH:mm` when > 24hrs ago
  - [ ] Show tooltip on hover for absolute time
- [ ] LN invoices:
  - [x] Show invoices as expired
  - [ ] Show description and destination (Use Amboss or another source to show alias)
  - [ ] Show QR code first instead of directly opening invoice on desktop
- [ ] Cache:
  - [ ] Use IndexedDB for profile cache so that its searchable (currently only caching in `localStorage`)

## Bugs
- [x] Profile editing still has some cache bugs which make the UX bad
[...]

PR #3, opened 2023-01-05
Add global tab to Root

p2pseed:
Adding global tab to root page (Home page)

I also disabled the `timeline-related` sub as its causing spam on relays, In #2 i think you mentioned that you are aware of this anyway..


v0l (approved):
Cool, thanks! Tabs look weird but we can change this later

PR #11, opened 2023-01-06
feat: add support for positive and negative reactions

verbiricha:


src/element/Note.js
```diff
@@ -20,6 +20,8 @@ export default function Note(props) {
     const dataEvent = props["data-ev"];
     const reactions = props.reactions;
     const deletion = props.deletion;
+    const likes = reactions?.filter(({ Content }) => Content === "+").length ?? 0
```
v0l, line 23: Not all likes use `+` for content, reactions vary a lot between clients, on Damus they use 🤙

verbiricha: What do you think is best, to have emoji reactions or like/dislike? Added support for 🤙

v0l: I guess it would be pretty difficult to identify what is positive and what is negative outside of `+` & `-` since those are the only ones mentioned in the NIP, maybe its better not to distinguish them and to just show all reactions by their `Content` value https://github.com/nostr-protocol/nips/blob/master/25.md

Maybe we can just map well known contents into `+` and the rest are just generic reactions:
https://github.com/v0l/snort/blob/921122bd89810d9e300f01effd0aaaea23b36374/src/element/NoteReaction.js#L15-L26

verbiricha:
@v0l gotta work on having reactions data available elsewhere but this is looking good
Screenshot 2023-01-06 at 23 10 48

v0l (approved):
This is likely to change in the future but probably fine for now!

PR #13, opened 2023-01-06
feat: embed youtube videos

verbiricha:


src/Text.js
```diff
@@ -10,8 +10,10 @@ export function extractLinks(fragments) {
             return f.split(UrlRegex).map(a => {
                 if (a.startsWith("http")) {
                     try {
-                        let url = new URL(a);
-                        let ext = url.pathname.toLowerCase().match(FileExtensionRegex);
+                        const url = new URL(a);
+                        const vParam = url.searchParams.get('v')
+                        const isYoutube = (url.host === "www.youtube.com" || url.host === "youtube.com" ) && vParam
```
v0l, line 15: I would make this a separate function to iterate over fragments and have another Regex to parse the youtube links, can see this being the case for other sources in the future too

v0l:
We could also add some breaks around this: https://ab0a921a.snort-social.pages.dev/e/note1mju85jc587326vqgjkkn9f4sy5c4a578azayzhy4q0f54m4qslusfehluj

v0l:
Also styling should be extended for youtube iframes 
https://github.com/v0l/snort/blob/7831bdbcc28a46eea436afa32d0db252df200c13/src/element/Note.css#L25-L28

verbiricha:
@v0l addressed your comments, I think is best to transform HTTP links in one pass, added a better regex for extracting youtube vid ids.

PR #21, opened 2023-01-09
feat: nip05 on profile page

verbiricha:
Noticed cash.app doesn't set CORS so those don't show as verified but in general the nostr.json files are CORS-ready

Screenshot 2023-01-09 at 19 42 31

### DIFF
--- a/src/element/Nip05.js
+++ b/src/element/Nip05.js
@@ -1,0 +1,43 @@
+import { useState, useEffect } from "react";
+
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faCheck } from "@fortawesome/free-solid-svg-icons";
+
+import './Nip05.css'
+
+const Nip05 = ({ nip05, pubkey }) => {
+    const [nip05pubkey, setNip05pubkey] = useState()
+    const isVerified = nip05pubkey === pubkey
+    const [nick, domain] = nip05.split('@')
+
+    useEffect(() => {
+        fetch(`https://${domain}/.well-known/nostr.json`)
+          .then((res) => res.json())
+          .then(({ names }) => {
+            if (names && names[nick]) {
+              setNip05pubkey(names[nick])
+            }
+          })
+          .catch((err) => {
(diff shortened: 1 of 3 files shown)


src/element/Nip05.js
```diff
@@ -0,0 +1,43 @@
...
+
+import './Nip05.css'
+
+const Nip05 = ({ nip05, pubkey }) => {
+    const [nip05pubkey, setNip05pubkey] = useState()
+    const isVerified = nip05pubkey === pubkey
+    const [nick, domain] = nip05.split('@')
+
+    useEffect(() => {
+        fetch(`https://${domain}/.well-known/nostr.json`)
```
v0l, line 14: You should include the param `?name=${nick}` some services will require this

src/element/Nip05.js
```diff
@@ -0,0 +1,43 @@
...
+          .catch((err) => {
+            console.error("Couldn't verifiy nip05")
+          })
+    }, [nip05, nick, domain])
+
+    return (
+       <div className="flex nip05">
+         <div className="nick">{nick}</div>
+         <div className="domain">@{domain}</div>
+         {isVerified && (
```
v0l, line 30: If its not verified then we should show error right?

src/element/Nip05.js
v0l: https://github.com/nostr-protocol/nips/blob/master/05.md

verbiricha: Good point, done.

src/element/Nip05.js
verbiricha: Screenshot 2023-01-10 at 07 19 08

src/element/Nip05.js
verbiricha: Also treating `_` as the default user and following spec recommendation for display:
Screenshot 2023-01-10 at 07 18 49
Screenshot 2023-01-10 at 07 18 44

v0l (approved):

PR #23, opened 2023-01-10
Shows QR code first 

ivanacostarubio:
>  Show QR code first instead of directly opening invoice on desktop


v0l:
Hey, thanks for the PR, i think it might be better to reuse the LNURLTip popup and just shortcut to the invoice screen, just so we can keep those pay flows the same.

The Invoice widget got left behind after the LN tip button was added

v0l (approved):
Nice one thanks!

PR #31, opened 2023-01-11
feat: optimistically render user reactions on timeline

verbiricha:


src/element/Timeline.js
```diff
@@ -9,9 +11,12 @@ import NoteReaction from "./NoteReaction";
  */
 export default function Timeline({ global, pubkeys }) {
     const feed = useTimelineFeed(pubkeys, global);
+    const reactions = useSelector(s => s.reactions)
+    const feedReactions = feed?.others || []
+    const allReactions = [...new Set([...reactions.user, ...feedReactions])]
 
     function reaction(id, kind = EventKind.Reaction) {
-        return feed?.others?.filter(a => a.kind === kind && a.tags.some(b => b[0] === "e" && b[1] === id));
+        return allReactions.filter(a => a.kind === kind && a.tags.some(b => b[0] === "e" && b[1] === id)).map(Event.FromObject);
```
v0l, line 19: The mapping on the end isnt needed here, we only store the plain objects up to this point, you can see this because we use `data` prop on `Note` instead of `data-ev`

PR #53, opened 2023-01-13
feat: display banner in profile

verbiricha:


v0l (commented):
Looks great, would be cool to keep existing layout if there is no banner though? not really a fan of having the avatar as the default banner

v0l (approved):
Perfect!

PR #54, opened 2023-01-14
DM's

v0l:


verbiricha:
Still not clearing the input value for me after sending a DM, other than that lgtm :fire:

verbiricha:
Should we add a message icon on profiles to send DMs to a user? Can't seem to send to anyone who didn't message me.

verbiricha:
Looks great, only thing missing is an icon to start a DM conversation with a user but I can add that later when I do a UI pass.

verbiricha (approved):

Issue #47, opened 2023-01-12
Local searchable user cache on IndexedDB


verbiricha:
- [x] Update `useProfile` to use cached data
- [x] Deprecate localstorage user storage
- [ ] Query for autocompletion on Indexed DB

v0l:
#65

v0l:
Closed by #76

PR #56, opened 2023-01-14
feat: mentions

verbiricha:


src/element/Textarea.js
```diff
@@ -0,0 +1,55 @@
...
+
+import Nip05, { useIsVerified } from "./Nip05";
+import "@webscopeio/react-textarea-autocomplete/style.css";
+import "./Textarea.css";
+import Nostrich from "../nostrich.jpg";
+
+function searchUsers(query, users) {
+  const q = query.toLowerCase()
+  return Object.values(users).filter(({ name, display_name, about }) => {
+    return name.toLowerCase().includes(q) || display_name?.includes(q) || about?.includes(q)
```
v0l, line 13: nip5 includes

src/element/Textarea.js
```diff
@@ -0,0 +1,55 @@
+import { Component } from "react";
```
v0l, line 1: This should be `tsx` now, can you convert it to TS?

src/feed/EventPublisher.js
```diff
@@ -28,6 +30,25 @@ export default function useEventPublisher() {
         return ev;
     }
 
+
+    function processMentions(ev, msg) {
+        const replaceHexKey = (match) => {
```
v0l, line 35: Trying to remove hex keys i think, only `npub` / `note` keys should be extracted

verbiricha: Leaving note references out for now, this PR only adds pubkey mentions with bech32 notation.

verbiricha:
Should be much better after https://github.com/v0l/snort/issues/47 but this is a start.

v0l (approved):

PR #65, opened 2023-01-15
add user DB and cache nip-05 verifications

verbiricha:


src/element/Textarea.tsx
```diff
@@ -37,29 +39,39 @@ const UserItem = ({ pubkey, display_name, picture, nip05, ...rest }: User) => {
...
+function normalizeUser({ pubkey, about, nip05, name, display_name }: User) {
+  return { pubkey, about, nip05, name, display_name }
+}
+
+const Textarea = ({ users, onChange, ...rest }: any) => {
+    const normalizedUsers = Object.keys(users).reduce((acc, pk) => {
+      return {...acc, [pk]: normalizeUser(users[pk]) }
+    }, {})
+    const dbUsers = useLiveQuery(
+      () => db.users.toArray().then(usrs => {
```
v0l, line 51: We can query directly inside here instead of cloning the entire cache?

src/state/Users.js
```diff
@@ -67,6 +68,14 @@ const UsersSlice = createSlice({
...
                 state.users[x.pubkey] = x;
+                db.users.put({
+                  pubkey: x.pubkey,
+                  name: x.name,
+                  display_name: x.display_name,
+                  nip05: x.nip05,
+                  picture: x.picture,
+                  about: x.about,
+                })
                 window.localStorage.setItem(`user:${x.pubkey}`, JSON.stringify(x));
```
v0l, line 79: Not needed anymore right?

src/element/Textarea.tsx
verbiricha: Yes I need to do that.

v0l (approved):

PR #83, opened 2023-01-18
autocomplete improvements

verbiricha:
- match npubs
- emojis :shipit:


src/db/index.ts
```diff
@@ -1,14 +1,19 @@
...
 
 
 export class SnortDB extends Dexie {
   users!: Table<MetadataCache>;
 
   constructor() {
     super('snortDB');
-    this.version(1).stores({
-      users: '++pubkey, name, display_name, picture, nip05' // Primary key and indexed props
+    this.version(2).stores({
```
v0l, line 11: Do you need to keep the v1 specification?

v0l (approved):

src/db/index.ts
verbiricha: not needed.

Issue #84, opened 2023-01-18
markdown <li> error


Saibato:
handling of  \<li\> leads to an error from react-markdown lib

i..e see thread
note1xpy0ggxlawptv585sfwzzrhhnw0qva9h6xk0dgkcvzkn0m9ngz3s7684yf
http://snort.social/e/note1xpy0ggxlawptv585sfwzzrhhnw0qva9h6xk0dgkcvzkn0m9ngz3s7684yf

![Screenshot from 2023-01-18 22-19-14](https://user-images.githubusercontent.com/34917548/213308094-0b40b91d-d87f-4261-a38c-260657e9cba4.png)

PR #85, opened 2023-01-18
bug fix: remove components handling of <li> by react markdown

Saibato:
possible fix to: [#84](https://github.com/v0l/snort/issues/84#issue-1548221905)

react markdown seam to crash if \<li\> is also handled in components

### DIFF
--- a/src/element/Text.tsx
+++ b/src/element/Text.tsx
@@ -142,11 +142,6 @@ function extractHashtags(fragments: Fragment[]) {
     }).flat();
 }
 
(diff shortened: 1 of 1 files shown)


v0l:
Fixed in 36fc7715353e38db020e91500d21961997cff65e

Saibato:
> Fixed in [36fc771](https://github.com/v0l/snort/commit/36fc7715353e38db020e91500d21961997cff65e)

yup that fix works way better, test confirmed on andriod , :1st_place_medal:  thanks

Issue #39, opened 2023-01-12
Background notifications


verbiricha:
- [x] Service worker with foreground notifications
- [ ] Background notifications worker

PR #88, opened 2023-01-19
Notifications

v0l:
Add some basic notifications, also allows and prompts to install as a PWA as we have registered a service worker now

Ref #39 
image


src/feed/LoginFeed.ts
```diff
@@ -61,13 +64,41 @@ export default function useLoginFeed() {
...
+    return null;
+}
+
+async function sendNotification(ev: TaggedRawEvent) {
+    let n = await makeNotification(ev);
+    if (n != null && Notification.permission === "granted") {
+        let worker = await navigator.serviceWorker.ready;
+        worker.showNotification(n.title, {
+            body: n.body,
+            icon: Nostrich,
```
verbiricha, line 98: Can we set the `icon` to the picture to whoever wrote the note?

src/feed/LoginFeed.ts
```diff
@@ -61,13 +64,41 @@ export default function useLoginFeed() {
...
+    }
+    return null;
+}
+
+async function sendNotification(ev: TaggedRawEvent) {
+    let n = await makeNotification(ev);
+    if (n != null && Notification.permission === "granted") {
+        let worker = await navigator.serviceWorker.ready;
+        worker.showNotification(n.title, {
+            body: n.body,
```
verbiricha, line 97: We need to at least extract mentions to #e and #p in body or they appear as raw `#[0]` index notation

v0l: image

verbiricha (approved):

PR #93, opened 2023-01-19
first pass at nip42 authentication

LiranCohen:
This is a first pass at implementing Nip42 Authentication.

I wasn't really sure where to put the auth function so I put it into the publisher since it has all of that type of logic there.
But I didn't really know how to get the publisher (or the event signing) into where the connection and AUTH message is handled.

There is probably a much cleaner way, maybe creating an event listener?

I figured I'd put down some code and test things out against my NIP42 relay and get some feedback on how you'd best want to accomplish this.


v0l:
What relays support auth?

LiranCohen:
I don't see any of the current public relays I chat on support it, but I've been coding on:
https://github.com/fiatjaf/relayer

It currently uses NIP42 as a guard around querying for Kind 4.

Can put this off 'til later if you don't see a need for it.

Will be happy to re-implement when that time comes.

verbiricha:
@LiranCohen what is your npub? want to follow you on Nostr.

v0l:
I think we can implement this into the snort relay i like the idea, as well as the search nip

LiranCohen:
@verbiricha I'm `npub148jmlutaa49y5wl5mcll003ftj59v79vf7wuv3apcwpf75hx22vs7kk9ay` 

I'll have some time this weekend to try and re-do this in a cleaner way.

src/Nostr/Connection.ts
```diff
@@ -94,17 +99,11 @@ export default class Connection {
...
-        for (let p of this.Pending) {
-            this._SendJson(p);
-        }
-        this.Pending = [];
-
-        for (let [_, s] of this.Subscriptions) {
-            this._SendSubscription(s);
-        }
-        this._UpdateState();
+        setTimeout(() => {
```
LiranCohen, line 102: I did some tests and it REALLY doesn't like this timeout.

That's fine the subscriptions will just be re-requested on auth, which happens. I'll do some further testing.

src/Feed/EventPublisher.ts
```diff
@@ -77,6 +79,17 @@ export default function useEventPublisher() {
...
+        nip42Auth: async (challenge: string, relay:string): Promise<NEvent> => {
+            if(pubKey) {
+                const ev = NEvent.ForPubKey(pubKey);
+                ev.Kind = EventKind.Auth;
+                ev.Content = "";
+                ev.Tags.push(new Tag(["relay", relay], 0));
+                ev.Tags.push(new Tag(["challenge", challenge], 1));
+                return await signEvent(ev);
+            }
+            return Promise.reject();
```
v0l, line 91: You can simply just not return anything here instead, undefined is fine

src/Pages/Layout.tsx
```diff
@@ -24,10 +26,20 @@ export default function Layout() {
     const readNotifications = useSelector<RootState, number>(s => s.login.readNotifications);
     const dms = useSelector<RootState, RawEvent[]>(s => s.login.dms);
     const prefs = useSelector<RootState, UserPreferences>(s => s.login.preferences);
+    const pub = useEventPublisher();
     useLoginFeed();
 
     useEffect(() => {
         if (relays) {
+            const nip42AuthEvent = async (event: NIP42AuthChallenge) => {
```
v0l, line 34: I think it would be simpler to just attach a function to `System` at startup which can handle these auth methods, instead of using the window event listener

v0l (commented):
`nostr-rs-relay` recently added support for this so it should be coming soon

src/Pages/Layout.tsx
LiranCohen: Yeah, that does seem cleaner.

I just made some changes and pushed.

LiranCohen:
@v0l made the suggested changes, definitely is cleaner than passing events to the window.

v0l (approved):

PR #95, opened 2023-01-19
Tidal embeds

v0l:
telegram-cloud-photo-size-4-5929449167174875946-y

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -8,7 +8,7 @@
   <meta name="theme-color" content="#000000" />
   <meta name="description" content="Fast nostr web ui" />
   <meta http-equiv="Content-Security-Policy"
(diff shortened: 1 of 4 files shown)


src/element/TidalEmbed.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+    const ScriptSrc = "https://embed.tidal.com/tidal-embed.js";
+    useEffect(() => {
+        let sTag = document.createElement("script");
+        sTag.src = ScriptSrc;
+        sTag.async = true;
+        document.head.appendChild(sTag);
+    }, []);
+
+    if (!data) return null;
```
verbiricha, line 23: shouldn't we fall back to rendering the link if we can embed it?

verbiricha: **can't**

verbiricha (approved):

PR #105, opened 2023-01-20
Add note context menu

v0l:
image

### DIFF
--- a/src/Element/Note.css
+++ b/src/Element/Note.css
@@ -1,43 +1,55 @@
 .note.thread {
-    border-bottom: none;
+  border-bottom: none;
 }
 
-.note > .header .reply {
-    font-size: var(--font-size-tiny);
-    color: var(--font-secondary-color);
+.note>.header .reply {
+  font-size: var(--font-size-tiny);
+  color: var(--font-secondary-color);
 }
 
-.note > .header > .info {
-    font-size: var(--font-size);
-    white-space: nowrap;
-    color: var(--font-secondary-color);
+.note>.header>.info {
+  font-size: var(--font-size);
+  white-space: nowrap;
+  color: var(--font-secondary-color);
 }
 
-.note > .body {
-    margin-top: 12px;
-    padding-left: 56px;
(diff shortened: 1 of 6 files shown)


src/Element/NoteFooter.tsx
```diff
@@ -111,27 +112,65 @@ export default function NoteFooter(props: NoteFooterProps) {
...
+        title: "Snort",
+        url: url
+      });
+    } else {
+      await navigator.clipboard.writeText(url);
+    }
+  }
+
+  function menuItems() {
+    return (
```
verbiricha, line 141: When we add note mention we could add another option here for copying the raw note id so you can paste it in a note for mentioning it.

src/Element/NoteFooter.tsx
```diff
@@ -140,15 +179,16 @@ export default function NoteFooter(props: NoteFooterProps) {
             <FontAwesomeIcon icon={faReply} />
           </div>
         </div>
+        <Menu menuButton={<div className="reaction-pill">
```
verbiricha, line 182: Screenshot 2023-01-21 at 09 06 59

wdyt about removing a bit of x padding? looks a bit wide.

verbiricha (approved):
minor comments, lgtm :shipit:

src/Element/Note.css
```diff
@@ -1,43 +1,55 @@
...
+  border-radius: 16px;
+}
+
+.note>.header img:hover, .note>.header .name>.reply:hover, .note .body:hover {
+  cursor: pointer;
 }
 
-.note > .note-creator {
+.note>.note-creator {
   margin-top: 12px;
```
verbiricha, line 42: I think we should pad this so is aligned with note body.

```suggestion
  margin-top: 12px;
  padding-left: 56px;
```

src/Element/NoteFooter.tsx
v0l: image

src/Element/Note.css
v0l: image

PR #110, opened 2023-01-21
feat: in-memory fallback for storing user profiles

### LINKED ISSUE #72: Make sure we handle DB write errors
In Firefox private mode:

```
{
  "_e": {},
  "name": "DatabaseClosedError",
  "message": "InvalidStateError A mutation operation was attempted on a database that did not allow mutations.\n InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.",
  "inner": {
    "_e": {},
    "name": "InvalidStateError",
    "message": "A mutation operation was attempted on a database that did not allow mutations.\n InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.",
    "inner": {}
  }
}
```

verbiricha:
Visit https://dbfix.snort-social.pages.dev/ with Tor

resolves #72


src/State/Users.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+});
+
+const { setUsers } = UsersSlice.actions
+
+function groupByPubkey(acc: Record<HexKey, MetadataCache>, user: MetadataCache) {
+  return { ...acc, [user.pubkey]: user }
+}
+
+export const add = async (user: MetadataCache) => {
```
v0l, line 69: This is great, i think we can improve it one more way, if we create an interface which defines these methods, then instead of using try/catch with a `const SnortDb` we can have `const db = SnortDb | ReduxDb` which both implement the users store, no need for try/catch then

At startup we can test if indexdDb is available with `window.indexdDb.open("test").catch(() => "Not supported")` or something similar, at this point we initialize the `db` to be `ReduxDb` which implements `UsersDb` or some other interface

v0l: Hey if you dont have time to make these changes maybe lets just merge it now and work on it later

src/State/Users/Hooks.ts
```diff
@@ -0,0 +1,51 @@
...
+  )
+
+  return allUsers
+}
+
+export function useKey(pubKey: HexKey) {
+  const user = useLiveQuery(async () => {
+      if (pubKey) {
+        try {
+          return await indexedDb.find(pubKey);
```
v0l, line 24: `db.find`?

src/State/Users/Db.ts
```diff
@@ -0,0 +1,147 @@
...
+import { db as idb } from "Db";
+
+import { UsersDb, MetadataCache, setUsers } from "State/Users";
+import store from "State/Store";
+
+class IndexedDb implements UsersDb {
+  isAvailable() {
+    if ("indexedDB" in window) {
+      return new Promise<boolean>((resolve) => {
+        const req = window.indexedDB.open("test", 1)
```
v0l, line 11: Might as well try to open the SnortDb so we dont have a random "test" db be there forever

v0l (commented):
There seems to be some subscription spam now on global, need to be careful about how we modify state to prevent rebuilding feeds 
image

v0l (approved):

src/State/Users/Db.ts
```diff
@@ -0,0 +1,147 @@
...
+import { db as idb, NAME, VERSION } from "Db";
+
+import { UsersDb, MetadataCache, setUsers } from "State/Users";
+import store from "State/Store";
+
+class IndexedDb implements UsersDb {
+  isAvailable() {
+    if ("indexedDB" in window) {
+      return new Promise<boolean>((resolve) => {
+        const req = window.indexedDB.open(NAME, VERSION)
```
v0l, line 11: This seems to create an empty database and not store anything inside?

verbiricha: I'll go back to using a dummy `test` db with version 1 then, wdyt?

v0l: Still doesnt seem to actually use IndexdDb, in logs it says its using indexdb but its not storing anything there

verbiricha: My bad, https://github.com/v0l/snort/pull/110/commits/59878776d5a377e62937d1120c3b73a011a91c95 should fix it

Issue #98, opened 2023-01-19
Soundcloud embed


verbiricha:

ivanacostarubio:
Should not render a player 
https://snort.social/e/2ee5cbc3bad838460a039cc91d7e1b01ccc26ad0f29d40978598890943798bc5 

Should render a player 
https://snort.social/e/note15c4d8yef69drhqc0vujmp5y0nvu5nkyev4v4xxp3h0qujcxqx9aqf86klu

PR #112, opened 2023-01-21
feat: soundcloud embed

ivanacostarubio:
https://github.com/v0l/snort/issues/98

### DIFF
--- a/src/Element/SoundCloudEmded.tsx
+++ b/src/Element/SoundCloudEmded.tsx
@@ -1,7 +1,6 @@
 const SoundCloudEmbed =  ({link}:  {link: string}) => {
 
     return(
-        <>
         <iframe
             width="100%"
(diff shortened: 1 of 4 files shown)


src/Element/SoundCloudEmded.tsx
```diff
@@ -0,0 +1,16 @@
+const SoundCloudEmbed =  ({link}:  {link: string}) => {
+
+    return(
+        <>
```
v0l, line 4: Fragment is not needed if its a single element

ivanacostarubio: https://github.com/v0l/snort/pull/112/commits/ea61b01f375a6fbd7b29ebc4af6306555ceb07d0

v0l (approved):

PR #131, opened 2023-01-24
feat: keyword search draft

brugeman:
My first react code, so be prepared.
It doesn't update the search input text (not sure how to do that), and I had to hard-code my relay as the 'search relay'. When the NIP is approved and some other relays implement this it could be generalized. Or maybe if not many relays support it then we could add a 'search relays' setting to let people have control.


v0l:
Hey thanks for the PR, actually looks fine, can you delete the `package-lock` file? We dont use NPM we use yarn

You can install deps with `yarn install`

brugeman:
> Hey thanks for the PR, actually looks fine, can you delete the `package-lock` file? We dont use NPM we use yarn
> 
> You can install deps with `yarn install`

Done.

v0l (approved):

PR #133, opened 2023-01-24
UI Changes for Profile & Preferences

FlannelDipole:
**Changed**
- Message reactions -> "Message reaction options" and made it more clear what it does
- "System (Default)" -> "System" to follow general UI standard

**Added**
- " Buy NIP-05" instead of just "Buy" to clarify what exactly a person is buying

### DIFF
--- a/src/Pages/settings/Preferences.tsx
+++ b/src/Pages/settings/Preferences.tsx
@@ -17,7 +17,7 @@ const PreferencesPage = () => {
                 </div>
                 <div>
                     <select value={perf.theme} onChange={e => dispatch(setPreferences({ ...perf, theme: e.target.value} as UserPreferences))}>
-                        <option value="system">System (Default)</option>
+                        <option value="system">System</option>
                         <option value="light">Light</option>
                         <option value="dark">Dark</option>
                     </select>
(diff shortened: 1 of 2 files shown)


src/Pages/settings/Preferences.tsx
```diff
@@ -34,8 +34,8 @@ const PreferencesPage = () => {
             </div>
             <div className="card flex">
                 <div className="flex f-col f-grow">
-                    <div>Enable reactions</div>
-                    <small>Reactions will be shown on every page, if disabled no reactions will be shown</small>
+                    <div>Message reaction options</div>
```
v0l, line 37: "options" is plural but this is a single checkbox, i think "Enable Reactions" was better

src/Pages/settings/Preferences.tsx
```diff
@@ -17,7 +17,7 @@ const PreferencesPage = () => {
                 </div>
                 <div>
                     <select value={perf.theme} onChange={e => dispatch(setPreferences({ ...perf, theme: e.target.value} as UserPreferences))}>
-                        <option value="system">System (Default)</option>
+                        <option value="system">System</option>
```
v0l, line 20: Default refers to it being the default option, not the systems default theme

src/Pages/settings/Profile.tsx
```diff
@@ -133,7 +133,7 @@ export default function ProfileSettings() {
                         <div className="btn" onClick={() => navigate("/verification")}>
                             <FontAwesomeIcon icon={faShop} />
                             &nbsp;
-                            Buy
+                            Buy NIP-05
```
v0l, line 136: Buy button is inline with NIP-05 entry i think its clear enough what you are buying

PR #135, opened 2023-01-24
UI updates

verbiricha:
Not changed:
- Threads (designs wip, will tackle separately)
- Direct messages (no designs)
- Settings (no designs)


v0l:
1. Login / Profile editor buttons still use old styles
2. Note creator clips under keyboard on android
3. Banner missing fade out

image
image
image

v0l:
It seems like we're changing the design every few days lol, maybe we can do entire UI in one go so that it at least remains consistent for more than a few days, that said i do really like the new designs!

verbiricha:
> It seems like we're changing the design every few days lol, maybe we can do entire UI in one go so that it at least remains consistent for more than a few days, that said i do really like the new designs!

😂 Yeah I'm going to do all in this PR, still work in progress

verbiricha:
@v0l we still don't have designs for the screens i listed above, perhaps we should ship this and then improve those when we have the designs.

v0l (approved):

PR #142, opened 2023-01-25
Dim font color

Marfusios:
The font is too white. 
Proposed value is copied from Github dark mode.

image


verbiricha:
nothing wrong with white font imo

bitcoinuser:
> The font is too white.

I agree.

v0l:
I wouldn't be against dimming the font a bit, can be a bit harsh sometimes

PR #143, opened 2023-01-26
Search

v0l:
Preview build: https://search.snort-social.pages.dev/

Todo:
- [x] Use any relays which support [NIP-50](https://github.com/nostr-protocol/nips/blob/master/50.md)
- [x] Connect only to default search relays when entering search page
- [x] Change search input to be link to base search page `/search` where user can enter term (this makes it easier to manage connections to default search relays)


verbiricha:
Can we also add kind 0 metadata to the search? Would be cool to display users too. Should we exclude reposts from search?

v0l:
> Can we also add kind 0 metadata to the search? Would be cool to display users too. Should we exclude reposts from search?

I suppose it depends on what is indexed, right now we just query for text note, maybe its also possible to query kind0 ? @brugeman

PR #144, opened 2023-01-26
Nip42 (AUTH)

### LINKED ISSUE #94: NIP42 Support
I'd like to get NIP42 support within SNORT.
https://github.com/nostr-protocol/nips/blob/master/42.md

I started a PR to get some of the code in there, not really sure I love where I placed everything but wanted to get some conversation going.

https://github.com/v0l/snort/pull/93

v0l:
Seems to work fine, tested with 2 different relay implementations,

`nostr-rs-relay` and `nostream` are the 2 most popular relay softwares, we should also make sure to test it on these before merging

https://nip42.snort-social.pages.dev/ 

Closes #94


LiranCohen:
I tested it out a bit on my private relay, seems to be working well.

I noticed that it tries to AUTH against a server that is not marked to be read from, which I think it shouldn't do.

I noticed this because I *thought* my server had read settings set to true, but I only saw it AUTH successfully and not send any other REQ messages (or receive events).

https://github.com/v0l/snort/pull/145

src/Pages/Layout.tsx
```diff
@@ -24,8 +25,13 @@ export default function Layout() {
     const readNotifications = useSelector<RootState, number>(s => s.login.readNotifications);
     const dms = useSelector<RootState, RawEvent[]>(s => s.login.dms);
     const prefs = useSelector<RootState, UserPreferences>(s => s.login.preferences);
+    const pub = useEventPublisher();
     useLoginFeed();
 
+    useMemo(() => {
```
verbiricha, line 31: we are not using result so can use `useEffect`

```suggestion
    useEffect(() => {
```

verbiricha (approved):
[Nostream](https://github.com/Cameri/nostream) doesn't seem to support NIP-42 according to README, neither does [nostr-rs-relay](https://github.com/scsibug/nostr-rs-relay). We should test with them when they release NIP-42 support, code lgtm.

v0l:
> [Nostream](https://github.com/Cameri/nostream) doesn't seem to support NIP-42 according to README, neither does [nostr-rs-relay](https://github.com/scsibug/nostr-rs-relay). We should test with them when they release NIP-42 support, code lgtm.

The good thing about this is, it doesnt matter if the relay supports it or not, so we can merge this anyway it will only respond to AUTH if the relay asks for it

LiranCohen:
🎉

PR #145, opened 2023-01-26
don't auth if settings are not set to read

LiranCohen:


src/Nostr/Connection.ts
```diff
@@ -352,6 +352,10 @@ export default class Connection {
     }
 
     async _OnAuthAsync(challenge: string): Promise<void> {
+        if(!this.Settings.read) {
```
v0l, line 355: I think we can always auth, because if you want to write an event but AUTH is required then it wont work

LiranCohen: Hrm, I wonder if that's confusing. 

I only noticed this because I was debugging and didn't realize my private relay was set to read = false, so after auth I wasn't getting any events.

I initially thought it didn't matter, but then realized writing is usually self-authed (since you're signing the message) but reading is where the NIP42 really matters.

I don't mind either way, just caught this as I was testing and it made me think a bit.

Can close if you think it can auth anyway, it won't hurt performance or anything (it's just a single message).

v0l: Well the server doesn't know that you signed the event, that's the entire point of NIP42, its to prove that you own the key (DM'S)

PR #146, opened 2023-01-26
feat: audio player

verbiricha:


v0l:
On void.cat we use these media types: 
```
case "audio/aac":
                case "audio/opus":
                case "audio/wav":
                case "audio/webm":
                case "audio/midi":
                case "audio/mpeg":
                case "audio/ogg": {
                ```

v0l (approved):

PR #151, opened 2023-01-27
Muted list via NIP-51

### LINKED ISSUE #101: Use Lists for storting block/mute lists
- [NIP PR](https://github.com/nostr-protocol/nips/pull/114)

Lists can be used to store public or private lists of note ids or pubkeys. The immediate and most important use case is encrypted block/mute lists that are interoperable between clients but we can explore creating lists to follow `e`, `t` and `p` tags in the future.

verbiricha:
Wanted to get some feedback on this, I think having a public mute list and the possibility to batch mute people is good, I can see who you've muted and preemptively mute them, wdyt?

I'm going to add a "Block" option (private mute) that stores the list of blocked `p` in the encrypted content field so we get public/private blocks. Not going to tackle other types of lists in this PR, just focused on mute list for `p` which I think is the most valuable atm. We can add pins and bookmarks later.

Closes #101

- [x] Mute (public)
- [x] Block (private)


verbiricha:
Link to NIP PR: https://github.com/nostr-protocol/nips/pull/183

src/Feed/MuteList.ts
```diff
@@ -0,0 +1,50 @@
...
+    return useSubscription(sub);
+}
+
+export function getMutedKeys(rawNotes: TaggedRawEvent[]): { createdAt: number, keys: HexKey[] } {
+    const notes = [...rawNotes]
+    notes.sort((a, b) => a.created_at - b.created_at)
+    const newest = notes && notes[0]
+    if (newest) {
+        const { tags } = newest
+        const mutedIndex = tags.findIndex(t => t[0] === "d" && t[1] === MUTE_LIST_TAG)
```
v0l, line 36: I dont think the index is important, you can just pick all `p` tags

v0l (commented):
Small bits, looks good though

src/Element/MutedList.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useSelector } from "react-redux";
+
+import { HexKey } from "Nostr"; import type { RootState } from "State/Store";
+import MuteButton from "Element/MuteButton";
+import ProfilePreview from "Element/ProfilePreview";
+import useMutedFeed, { getMuted } from "Feed/MuteList";
+import useModeration from "Hooks/useModeration";
+
+export interface MutedListProps {
+    pubkey: HexKey
```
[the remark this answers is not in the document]
verbiricha: Change this a bit, the `MutedList` component will show public mutes for users different that the logged in user. For logged in user added `BlockList` which displays public (mute) and private (block) muted.

src/Feed/LoginFeed.ts
```diff
@@ -108,6 +126,11 @@ export default function useLoginFeed() {
...
     }, [notificationFeed.store]);
 
+    useEffect(() => {
+      const ps = getMutedKeys(mutedFeed.store.notes)
+      dispatch(setMuted(ps))
+    }, [mutedFeed.store])
+
     useEffect(() => {
         let dms = dmsFeed.store.notes.filter(a => a.kind === EventKind.DirectMessage);
         dispatch(addDirectMessage(dms));
```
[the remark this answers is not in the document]
verbiricha: I've decided against it since we don't always have `muted` list available by the time we start getting dm events. Doing filtering on the downstream components instead.

src/Element/NoteReaction.tsx
```diff
@@ -53,8 +55,9 @@ export default function NoteReaction(props: NoteReactionProps) {
         showHeader: ev?.Kind === EventKind.Repost,
         showFooter: false,
     };
+    const isOpMuted = root && isMuted(root.pubkey)
 
-    return (
+    return isOpMuted ? null : (
```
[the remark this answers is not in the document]
verbiricha: For notes I added the `HiddenNote`, do you think is worth adding for reactions too? thought we could just hide them since they can get spammy

src/Feed/LoginFeed.ts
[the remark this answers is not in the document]
v0l: Good point

src/Element/NoteReaction.tsx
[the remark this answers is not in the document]
v0l: We dont show pure reactions currently, only reposts, im thinking of reposts where the person being reposted in muted, if the reposter is muted then you wont see it anyway because it will be removed at timeline/thread level

src/Element/BlockList.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const { blocked, muted } = useModeration();
+
+    return (
+        <div className="main-content">
+          <h3>Muted ({muted.length})</h3>
+          {muted.map(a => {
+            return <ProfilePreview actions={<MuteButton pubkey={a} />} pubkey={a} options={{ about: false }} key={a} />
+          })}
+          <h3>Blocked ({blocked.length})</h3>
+          {blocked.map(a => {
```
v0l, line 22: If you had a really long mute list then you would have to scroll for a long time to get to blocks, should we not have seperate tab for when its the logged in users block list?

src/Element/NoteReaction.tsx
```diff
@@ -18,6 +19,9 @@ export interface NoteReactionProps {
 export default function NoteReaction(props: NoteReactionProps) {
     const { ["data-ev"]: dataEv, data } = props;
     const ev = useMemo(() => dataEv || new NEvent(data), [data, dataEv])
+    const { isMuted } = useModeration();
+    const pRef = data?.tags.find((a: any) => a[0] === "p")?.at(1);
```
v0l, line 23: Better to use `ev.Tags` here because you have 2 ways to pass the event data

src/Element/NoteReaction.tsx
```diff
@@ -53,8 +57,9 @@ export default function NoteReaction(props: NoteReactionProps) {
         showHeader: ev?.Kind === EventKind.Repost,
         showFooter: false,
     };
+    const isOpMuted = root && isMuted(root.pubkey)
```
v0l, line 60: the `p` tag check and `root.pubkey` should be the same, do you think its needed to have both checks?

src/Feed/MuteList.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { HexKey, TaggedRawEvent, Lists } from "Nostr";
+import EventKind from "Nostr/EventKind";
+import { Subscriptions } from "Nostr/Subscriptions";
+import useSubscription, { NoteStore } from "Feed/Subscription";
+
+export default function useMutedFeed(pubkey: HexKey) {
+    const sub = useMemo(() => {
+        let sub = new Subscriptions();
+        sub.Id = `muted:${pubkey}`;
```
v0l, line 11: You can use only the start of the pubkey here, so the id is a bit snorter

v0l:
Closes #101

src/Element/NoteReaction.tsx
verbiricha: No, my bad

src/Feed/MuteList.ts
verbiricha: i see what you did there ✨

v0l:
When i click onto the muted persons profile it should show notes anyway?

verbiricha:
good point, changed it so when visiting Profile the moderation is ignored.

PR #153, opened 2023-01-27
Make Markdown more interoperable

fiatjaf:
By disallowing the annoying link/image syntax that would cause terrible rendering bugs in other clients.

As result of this change, this note:

```
# titles still work

- lists
- still
- work

normal https://literal.links and `monospace` still work

**bold** and _italic_ still work

just [annoying](https://syntax-heavy.links/likethis) do not work, and ![the even](https://more.annoying/images.png)

normal images work fine: https://pixy.org/src/487/4870083.jpg
```

Renders like this: 
2023-01-27-183610_699x896_scrot

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "react-textarea-autosize": "^8.4.0",
     "react-twitter-embed": "^4.0.4",
     "typescript": "^4.9.4",
+    "unist-util-visit": "^4.1.2",
     "uuid": "^9.0.0",
     "workbox-background-sync": "^6.4.2",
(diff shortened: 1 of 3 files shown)


verbiricha (commented):
I like the idea but one problem I see with this approach is that `(` and `)` are valid URL characters and we end up rendering broken URLs since `)` at the end is treated as part of the URL by our regex.

fiatjaf:
That is the point: to discourage people from posting stuff in this syntax. They are rendering broken like that on every other client too.

But sure, let me see if there is a good fallback for this.

verbiricha (approved):

PR #154, opened 2023-01-27
bug: logout reply

ivanacostarubio:


v0l (approved):

src/Element/NoteFooter.tsx
```diff
@@ -183,11 +183,13 @@ export default function NoteFooter(props: NoteFooterProps) {
   return (
     <>
       <div className="footer">
+        { login ?
         <div className={`reaction-pill ${reply ? 'reacted' : ''}`} onClick={(e) => setReply(s => !s)}>
           <div className="reaction-pill-icon">
             <FontAwesomeIcon icon={faReply} />
           </div>
         </div>
+        : "" }
```
v0l, line 192: Instead of `""` you can just do `{login && (<node>))}`

ivanacostarubio: https://github.com/v0l/snort/pull/154/commits/ea84a31a364b4fa78957132c57abb4acaa5dd667

v0l (approved):

PR #155, opened 2023-01-27
fix: default to in-memory db only on db read fail

verbiricha:
My bad, broke this by accident.


yarn.lock
```diff
@@ -6043,11 +6043,6 @@ js-sha3@0.8.0:
   resolved "https://registry.yarnpkg.com/js-sha3/-/js-sha3-0.8.0.tgz#b9b7a5da73afad7dedd0f8c463954cbde6818840"
   integrity sha512-gF1cRrHhIzNfToc802P800N8PpXS+evLLXfsVpowqmAFR9uwbi89WvXg2QspOmXL8QL86J4T1EpFu+yUkwJY3Q==
 
-js-stylesheet@0.0.1:
```
v0l, line 6046: I think you can remove these changes, its a mistake

verbiricha: for some reason these are removed automatiaclly when running yarn, not sure we need them but removing them anyway.

v0l (approved):

PR #161, opened 2023-01-28
New UI

v0l:


v0l:
Cant click the background to close the note creator

v0l:
The Note creator is also covered with on-screen keyboard on android

Instead of `align-content: center` we can just use `margin-top: 20vh` on <720px screen

verbiricha (approved):

Issue #159, opened 2023-01-28
Nostr protocol handler


v0l:
Register protocol handler for `nostr:` protocol so that links will open directly on Snort

https://github.com/nostr-protocol/nips/blob/master/21.md

Requires #158

Issue #158, opened 2023-01-28
Complete NIP-19 compat (TLV types)


v0l:
As a precursor to supporting `nostr:` protocol handler we need to complete support for TLV encoded links

https://github.com/nostr-protocol/nips/blob/master/19.md

PR #164, opened 2023-01-29
protocol handler

v0l:
Handle `nostr:` links

Ref #158 #159


verbiricha:
Should we use NIP-21 for shared URLs too?

verbiricha (approved):

v0l:
> Should we use NIP-21 for shared URLs too?

Seems like they wont be compatible, native apps can register `nostr:` but on web its restricted to `web+nostr:` so they wont even open on damus and damus links wont open on snort

Pretty shitty

PR #166, opened 2023-01-29
Show absolute time on hover

wanacode:
This is a suggestion to fix issue https://github.com/v0l/snort/issues/46

The NoteTime now renders as a `<time>` element and will show the absolute time on hover using the title attribute (similar to how GitHub does it).

I have also added a `datetime` attribute.


verbiricha (approved):
Thanks for doing this, lgtm ✨

PR #169, opened 2023-01-30
fix: send d tags as list

verbiricha:


v0l (changes requested):
Its not an array of tags, who said that? https://github.com/monlovesmango/nips/blob/lists/51.md

verbiricha:
mb, read it on telegram but they are wrong. closing this.

v0l:
Yea this is correct actually

v0l (approved):

PR #170, opened 2023-01-30
Threads

verbiricha:


verbiricha:
I'm going to tackle URL navigation separately, this can be reviewed already.

v0l:
Soem threads are hidden, maybe if im clicking into a note which is on 3rd level or deeper then it only shows the root and the note i click on is hidden somewhere: 

https://threads.snort-social.pages.dev/e/note1ucc2nvd0ta9xqxjje6rmqkds745hyeglv4hv0c0xvfac2yzs656sjcn06r

v0l:
Its still hiding this note, if i go to https://threads.snort-social.pages.dev/e/note1ucc2nvd0ta9xqxjje6rmqkds745hyeglv4hv0c0xvfac2yzs656sjcn06r it should show this note as focus? Right now i have to expand the Show more and there is no indication that its the selected note

verbiricha:
@v0l my bad, accidentaly removed the highlight. Added it back, also expand subthreads when active is there.

v0l:
Seems to not work still, when i navigate to post in a thread its highlighted at first but when the rest of the thread is loaded its buried

Instead of rendering the full thread we should skip to the level that the post is on and hide the upper levels?

verbiricha:
@v0l can you take another 👀 ?

v0l (approved):
❤️

PR #174, opened 2023-01-30
Image proxy service

v0l:
Added image proxy service which:
1. Converts avatars to webp
2. Resizes avatars to 200x200 (default)
3. Caches all avatars for 12hrs


verbiricha (approved):
✨

PR #190, opened 2023-02-02
Skeleton component on timeline loading for better user experience

leotuna:
**Before 👇**
skeleton_before

**After 👇**
skeleton_after

### DIFF
--- a/src/Element/LoadMore.tsx
+++ b/src/Element/LoadMore.tsx
@@ -1,7 +1,7 @@
 import { useEffect, useState } from "react";
 import { useInView } from "react-intersection-observer";
 
-export default function LoadMore({ onLoadMore, shouldLoadMore }: { onLoadMore: () => void, shouldLoadMore: boolean }) {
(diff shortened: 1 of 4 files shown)


verbiricha:
Hey Leo, thanks for this. One problem I see though is that is only looking good in light mode, for dark mode looks too bright:

Screenshot 2023-02-02 at 08 15 04

I also think it'd be great to give the skeletons the same border radius as the notes instead of being square.

leotuna:
What about now?
skeleton_2

Eiiki (commented):
Great initiative 🙌  
I would suggest not passing the `Skeleton` as a child to the `LoadMore` component and just render it directly from it instead. I think it's safe as `LoadMore` is only rendered from the `Timeline` component so its only purpose should be to render the `Skeleton`.

leotuna:
> Great initiative 🙌 I would suggest not passing the `Skeleton` as a child to the `LoadMore` component and just render it directly from it instead. I think it's safe as `LoadMore` is only rendered from the `Timeline` component so its only purpose should be to render the `Skeleton`.

Don't you think it should be reusable?

v0l (approved):
LGTM

PR #191, opened 2023-02-02
Use the current embed player via TIDALs OEmbed API.

enjikaka:
The code was using the old unmaintained embed player. This updates the code to use the newest one via our OEmbed API: https://oembed.tidal.com/?url=https://tidal.com/browse/track/145596276

The OEmbed API delivers the embed with sizes for different embed types. `dangerouslySetInnerHTML` is a bit sad, if wanted I can refactor to instead read out sizes from the delivered iframe and set those on a new <iframe> in the component.

https://user-images.githubusercontent.com/877455/216334497-f429daa3-4cf2-4dcd-8114-a0cabfa0fec3.mov

### DIFF
--- a/src/Element/TidalEmbed.tsx
+++ b/src/Element/TidalEmbed.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { CSSProperties, useEffect, useState } from "react";
 import { TidalRegex } from "Const";
 
 // Re-use dom parser across instances of TidalEmbed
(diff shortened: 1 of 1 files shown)


v0l:
Yea not a fan of injecting random HTML onto the page, why does this not work as an iframe?

v0l:
I mean why so the sizes need to be set on the iframes, can we not just set the width of the frame and the size is dynamic, that is what we do for spotify/youtube/soundcloud etc and it works fine

enjikaka:
Embeds from different services does not support a limitless amount of sizes. Their respective OEmbed API will report a supported recommended size. Dynamic heights aren't possible since the contents from an iframe can not affect the height of the iframe.

You can check the OEmbed spec here; all music-player embeds use the rich type -> https://oembed.com/#section2:~:text=2.3.4.4.-,The%20rich%20type,-This%20type%20is

enjikaka:
Updated now to use the source and height on a constructed iframe instead.

v0l (approved):
Really nice, thanks!

PR #193, opened 2023-02-03
Note creator improvements

verbiricha:


v0l:
I still cant really see note creator BG against background can we do something about that? @karnage

v0l (approved):

PR #205, opened 2023-02-06
fix Content Security Policy in index.html

middlingphys:
Currently, the `connect-src` defined in `index.html` only allows connections to the default port of the `wss:`/`https:` scheme, namely port 443.
This change will allow connections to a relay using the port other than 443.


middlingphys:
There are some relays with non-default ports, such as (taken from https://nostr.watch/relays)

- knostr.neutrine.com:8880
- deconomy-netser.ddns.net:2121

v0l:
Seems risky for very little reward

PR #207, opened 2023-02-06
fix: don't stream global feed in notifications tab

verbiricha:


src/Feed/TimelineFeed.ts
```diff
@@ -72,7 +73,7 @@ export default function useTimelineFeed(subject: TimelineSubject, options: Timel
                 }
             }
 
-            if (pref.autoShowLatest) {
+            if (canStream) {
                 // copy properties of main sub but with limit 0
```
v0l, line 77: Ah i see the issue:
```js
latestSub.PTags = sub.PTags
```

verbiricha: nice, pushed that.

v0l (approved):

Issue #127, opened 2023-01-24
Internationalization and localization


verbiricha:
Would be nice to support languages other than English. After supporting translatable strings and formatting we can set it up so it's easy for non-technical contributors to add translations to the app.

SamSamskies:
Any idea on lib you wanna use? I can setup the boilerplate if you wanna use `react-intl`.

v0l:
I just started reading i18next docs, is react-intl better?

SamSamskies:
I dunno. I've only used react-intl. Using react-intl with extract-react-intl-messages is pretty nice. Allows you to define the messages close to where they are used and it makes it pretty easy for non-technical people to add translations. 

I'll try and take a look at i18next today or tomorrow.

verbiricha:
any of those two is fine imo, would be nice to set up transifex for translations [like in Damus](https://github.com/damus-io/damus/blob/master/transifex.yml)

PR #216, opened 2023-02-08
react-intl spike

SamSamskies:
This possibly resolves https://github.com/v0l/snort/issues/127.

Not sure what languages you want to support. I started with just English and Spanish.

The locale is being set by whatever locale the user has set in their browser. You could also maybe let the user set the locale in their user settings or with a language selector on the landing page.

English
Screenshot 2023-02-07 at 11 47 00 PM

Spanish
Screenshot 2023-02-07 at 11 46 43 PM

### DIFF
--- a/src/Pages/messages.js
+++ b/src/Pages/messages.js
@@ -1,0 +1,8 @@
+import { defineMessages } from "react-intl";
+import { addIdAndDefaultMessageToMessages } from "Util";
+
+const messages = defineMessages({
+  Login: "Login",
+});
+
+export default addIdAndDefaultMessageToMessages(messages, 'Pages');
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,10 @@
+{
+  "plugins": [
+    [
+      "react-intl-auto",
+      {
+        "removePrefix": "src/"
+      }
+    ]
+  ]
+}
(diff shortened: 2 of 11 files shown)


src/Pages/messages.js
```diff
@@ -0,0 +1,8 @@
+import { defineMessages } from "react-intl";
+import { addIdAndDefaultMessageToMessages } from "Util";
+
+const messages = defineMessages({
+  Login: "Login",
+});
+
+export default addIdAndDefaultMessageToMessages(messages, 'Pages');
```
SamSamskies, line 8: This is example boilerplate file where default messages are defined. The values are extracted into one JSON file per locale support using the `yarn generate-messages` command.

The files are fairly simple, so I think it's okay to leave it them as JS.

verbiricha:
yo Sam thanks for starting this effort ❤️ 

>Not sure what languages you want to support. I started with just English and Spanish.

Starting with those is fine, once the project is ready for getting translations we'll probably publish it in Transifex to get translation contributions.

> The locale is being set by whatever locale the user has set in their browser. You could also maybe let the user set the locale in their user settings or with a language selector on the landing page.

Like this idea, i think this is a good default (system locale) but we should allow user to override that in the future, similar to theme selection.

verbiricha (approved):

v0l (commented):
Awesome, we really need to get Chinese and Japanese translations up first as those are the most popular

PR #220, opened 2023-02-08
translations

verbiricha:
I think I covered all except new user flow, I'm going to apply a bunch of design changes to that in an upcoming PR so leaving that out for now.


src/Pages/settings/messages.js
```diff
@@ -0,0 +1,60 @@
...
+  Name: "Name",
+  Profile: "Profile",
+  Website: "Website",
+  Save: "Save",
+  DisplayName: "Display name",
+  Buy: "Buy",
+  Nip05: "NIP-05",
+});
+
+export default addIdAndDefaultMessageToMessages(messages, "Settings");
```
SamSamskies, line 60: `Settings` should be lowercase cause it needs to match the pathname exactly. In this example, the path should be `Pages.settings`.

```js
export default addIdAndDefaultMessageToMessages(messages, "settings");
```

This helper util shouldn't be necessary, but I couldn't figure out how to inject the `id` and `defaultMessage` using the existing babel tooling. IMO this helper is good enough. Only problem is you have to make sure the prefix matches exactly what the path is. This could be improved by programmatically generating the prefix instead of hardcoding it like this.

SamSamskies: I'll look into programmatically generating the prefix.

v0l (approved):

PR #225, opened 2023-02-08
add ability to use babel plugins without ejecting

SamSamskies:
@verbiricha I figured out how to get babel-plugin-react-intl-auto to auto inject the message descriptors with babel.


verbiricha (approved):
awesome, needs to be rebased since we got japanese translations overnight 🤯

SamSamskies:
rebased 😀

PR #229, opened 2023-02-09
fix: allow zap comments

verbiricha:


src/Element/SendSats.tsx
```diff
@@ -77,6 +77,8 @@ export default function LNURLTip(props: LNURLTipProps) {
   const { formatMessage } = useIntl();
   const publisher = useEventPublisher();
   const horizontalScroll = useHorizontalScroll();
+  const canComment =
+    (payService?.commentAllowed ?? 0) > 0 || payService?.nostrPubkey;
```
v0l, line 81: Having a `nostrPubkey` doesnt mean you can pass comment, only `commentAllowed > 0`

src/Util.ts
```diff
@@ -108,24 +108,15 @@ export const Reaction = {
...
-      return Reaction.Positive;
-    case "🤙":
-      return Reaction.Positive;
-    case "❤️":
-      return Reaction.Positive;
-    case "👍":
-      return Reaction.Positive;
-    case "💯":
-      return Reaction.Positive;
     case "+":
```
v0l, line 111: Might as well also remove `+` and default to `Positive`

src/Element/SendSats.tsx
verbiricha: zaps do allow comments tho, they are part of the zap not the invoice comment afaict

v0l: Yes but this is for the `?comment=` param, not the comment of the zap note, i guess it doesn't matter if you pass comment arg and its not supported for zaps endpoint

v0l (approved):

src/Element/SendSats.tsx
verbiricha: will not send param if not supported by lnurl endpoint, good point

PR #235, opened 2023-02-09
feat: new login page

v0l:


src/Pages/Login.tsx
```diff
@@ -94,38 +136,88 @@ export default function LoginPage() {
...
-      <>
-        <h2>Other Login Methods</h2>
-        <div className="flex">
-          <button type="button" onClick={doNip07Login}>
-            Login with Extension (NIP-07)
-          </button>
-        </div>
-      </>
+      <button type="button" onClick={doNip07Login}>
+        Login with Extension (NIP-07)
```
verbiricha, line 140: this should also be translated, no?

src/Pages/Login.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.login .login-actions > button {
+  margin-right: 10px;
+}
+
+.login .login-or {
+  margin-top: 56px;
+  margin-bottom: 64px;
+}
```
verbiricha, line 83: I'd add a small margin to the left of the zap button:

```suggestion
}

.login .artwork .zap-button { 
  margin-left: 6px; 
}
```

verbiricha (approved):
minor comments, lgtm 🔥

src/Pages/Login.css
v0l: Didnt add margin, but removed the BG color so it looks like there is some spacing instead.

PR #238, opened 2023-02-09
NIP-65: Relay list metada

verbiricha:


src/Element/Relays.tsx
```diff
@@ -0,0 +1,47 @@
+import "./Relays.css";
+import Nostrich from "nostrich.webp";
+import { useState } from "react";
+
+import Read from "Icons/Read";
+import Write from "Icons/Write";
+
+export interface RelaySpec {
```
v0l, line 8: If this is part of NIP it should be in `Nostr/` dir?

src/Pages/settings/Relays.tsx
```diff
@@ -20,6 +20,8 @@ const RelaySettingsPage = () => {
     const ev = await publisher.saveRelays();
     publisher.broadcast(ev);
     publisher.broadcastForBootstrap(ev);
+    const settingsEv = await publisher.saveRelaysSettings();
```
v0l, line 23: Maybe we can write to all (or a lot of relays) from `https://api.nostr.watch/v1/online`

src/Element/Relays.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+interface RelaysProps {
+  relays: RelaySpec[];
+}
+
+const RelayFavicon = ({ url }: { url: string }) => {
+  const cleanUrl = url
+    .replace("wss://relay.", "https://")
+    .replace("wss://nostr.", "https://")
+    .replace("wss://", "https://")
```
v0l, line 21: `.replace("ws://", "http://")`

v0l (commented):
Maybe move `Relays.tsx` some other place since we already have `Relay.tsx`

src/Pages/settings/Relays.tsx
verbiricha: Sending to all of user relays and random sample of 20 from that endpoint.

v0l (approved):

PR #239, opened 2023-02-10
feat: reaction emoji

verbiricha:


src/State/Login.ts
```diff
@@ -176,6 +181,7 @@ export const InitState = {
   dmInteraction: 0,
   preferences: {
     enableReactions: true,
+    reactionEmoji: "+1",
```
v0l, line 184: Default should be `+`

verbiricha: my bad, was late when i did this 😹

PR #245, opened 2023-02-10
fix(banner): thinner

fernandolguevara:
before:

image

---

after:

image

### DIFF
--- a/src/Pages/ProfilePage.css
+++ b/src/Pages/ProfilePage.css
@@ -6,6 +6,7 @@
 
 .profile .banner {
   width: 100%;
+  border-radius: 5px;
   height: 160px;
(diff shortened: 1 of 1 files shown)


verbiricha:
Hey Fernando, thanks for the proposal but current sizes are based on the designs on https://www.figma.com/file/Q79i1WVA1bIAzYaLihgteN/Snort?node-id=55%3A102&t=liY0eV9bh8RQzPJX-0 and we are happy with how they look.

fernandolguevara:
@verbiricha, but it appears to not follow your Figma values.

image

v0l:
It was 280px height, not sure if we want to reduce height for different aspect ratio images.

PR #249, opened 2023-02-10
feat: read global from specific (paid) relays

v0l:


verbiricha:
Now if i'm logged out i can't see anything, also if I don't have paid relays Global doesn't show. Is it ok?

v0l:
I dont think we should show non-paid global, ill add some default paid relays for logged out

verbiricha:
Sounds good

PR #256, opened 2023-02-11
implement nip57 with suggested modifications

fiatjaf:
This implements modifications suggested on https://github.com/nostr-protocol/nips/pull/224#issuecomment-1418302436 while retaining backwards-compatibility (although I think it should _not_ retain that).

I wasn't able to test it yet because I don't have a compatible server (I was turning https://github.com/nbd-wtf/satdress into a zap-compatible server but realized that that will be very hard because currently it cannot know when an invoice was paid), so maybe it is broken.

For now, I'm leaving here just as an exercise.

### DIFF
--- a/src/Element/NoteFooter.tsx
+++ b/src/Element/NoteFooter.tsx
@@ -301,7 +301,8 @@ export default function NoteFooter(props: NoteFooterProps) {
           zaps={zaps}
         />
         <SendSats
-          svc={author?.lud16 || author?.lud06}
+          lnurl={author?.lud16 || author?.lud06}
+          nip57={author?.nip57}
           onClose={() => setTip(false)}
           show={tip}
           author={author?.pubkey}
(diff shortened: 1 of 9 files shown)


v0l:
I implemented the new zaps flow api already in snort donate, ill push it later to test

v0l:
Ok its live now, you can try use the new zaps flow here: https://api.snort.social/swagger/index.html

Payment status is only checked in background job so it will take about 2mins to see receipt

fiatjaf:
Thank you. I can only try this tomorrow, will proceed from there.

v0l:
> Thank you. I can only try this tomorrow, will proceed from there.

I already tried it out, i implemented it differently on the API

POST `/zap` returns something similar to LNURL endpoint:
```json
{ "pr": "lnbc1..." }
```

I also implement the same status response as LNURL

fiatjaf:
You mean the `status: "ERROR"` and the `reason` string? OK, that sounds good to me.

v0l:
Yep, pushed the changes, I got as far as making zaps but they are not showing in the UI

PR #274, opened 2023-02-11
Workspace with decoupled `nostr` package

ennmichael:
Turned the repo into a workspace. Decoupled `nostr` into its own package. Some ugliness involved but I will fix it in the following PRs.


packages/nostr/src/EventKind.ts
```diff
@@ -1,4 +1,4 @@
-const enum EventKind {
+enum EventKind {
```
ennmichael, line 1: `const enum` does not exist in JS and since this is a package now it has to export a JS interface.

v0l:
I tried to run `yarn start` in app but it failed, how should i run the site now?

ennmichael:
I updated the README:

> This repository is a yarn workspace. To install dependencies, run `yarn` from the project root. To run the application, use either
>
> ```sh
> $ yarn workspace @snort/app start
> ```
>
> or
>
> ```sh
> $ cd packages/app
> $ yarn start
> ```

ennmichael:
I also updated the `Dockerfile` so your deployments shouldn't break (though I'm not sure what you use for deployment).

v0l:
> I also updated the `Dockerfile` so your deployments shouldn't break (though I'm not sure what you use for deployment).

We use CF pages, basically clones the repo and runs `yarn build` will need to update that too

v0l (approved):

v0l (approved):

PR #286, opened 2023-02-13
HTML auto direction for specific textual content

verbiricha:


src/Pages/ProfilePage.tsx
```diff
@@ -150,10 +150,12 @@ export default function ProfilePage() {
 
   function bio() {
     return (
-      aboutText.length > 0 && (
-        <>
-          <div className="details">{about}</div>
-        </>
+      aboutText.length && (
+        <React.Fragment>
```
v0l, line 154: Fragment not needed

verbiricha:
updated the onboarding flow to use automatic lang direction and removed unneeded fragment.

v0l:
Is this the correct way to do it? What about english should that not be ltr?

v0l:
I mean it does seems to work: https://rtl.snort-social.pages.dev/e/note170sn3kcxxzk9unchk4zh7zatzxu7pjgkn59hyqpt4h0yg454m7vsdrh7q2

verbiricha:
auto will show text ltr or rtl depending on the content characters so does the right thing.

PR #291, opened 2023-02-13
SUPPORT RTL/LTR ON LOGIN PAGE

ahmedrowaihi:


src/Pages/Login.tsx
```diff
@@ -154,22 +154,23 @@ export default function LoginPage() {
           <div className="logo" onClick={() => navigate("/")}>
             Snort
           </div>
-          <h1>
+          <h1 dir="auto">
```
verbiricha, line 157: We can apply it to this whole container at once: `<div className="login-container" dir="auto">` and should work on all items.

ahmedrowaihi: I have tried but seems not applying to all elements.

You can test and confirm 👍

verbiricha: Interesting, wonder why that is. One minor nitpick: could we add `margin-right: 25px` after https://github.com/v0l/snort/blob/main/src/Pages/Login.css#L40 so it looks the same in rtl? Other than that lgmt, thanks for your contributions!

ahmedrowaihi: Sure, that's perfect.
Glad to contribute 🙏🏼.

ahmedrowaihi: Added @verbiricha

verbiricha (approved):

PR #299, opened 2023-02-14
Add reply count to thread view

joshr4:
I'm looking at trying to add the reply count next to the reply button while in the thread view.

I'm not too sure where to source the actual reply count number. I have found `brokenChains` and `chains`, but not sure what these two are, they are slightly different amounts for different notes. Can anyone shed some light on these two things? Should they be summed together?

Example screenshot of the number:
Screen Shot 2023-02-14 at 1 36 17 PM


v0l:
I dont think this is a good idea, we already have too much content being pulled from relays

Maybe when a `COUNT` NIP is accepted and relays support it we can do this

PR #300, opened 2023-02-14
Add build command to readme

joshr4:
With the recent yarn workspaces change (https://github.com/v0l/snort/pull/274), I was unable to initially run the app. I added some text to the readme that directs the user to run the build command for the `nostr` package so they can avoid `nostr` package errors.


SamSamskies (approved):
Thanks for the PR 🤙. I ran into this same problem today.

v0l:
I added a new script to the root, `yarn start` will work as normal

joshr4:
nice! https://github.com/v0l/snort/commit/94bdd44192337909a82dae6f3c0fe3c76359b7f7 adds `yarn start` which does the trick.

But, the two commands in the readme now only work if the user first builds the `nostr` package (via the root start or build scripts), should those two commands be removed in favor of suggesting the root `yarn start`? I updated this PR with one more commit 4bdc393029d1b8ec43debeac25ae685e260d6001

v0l (approved):

PR #301, opened 2023-02-14
UI bugs

verbiricha:


SamSamskies:
Formatting failed. I wonder if the pre-commit hook broke when converting the repo to a workspace.

SamSamskies:
I think the Prettier and Husky setup and maybe some other tooling configs needs to move to the root of the workspace.

v0l (approved):

Issue #177, opened 2023-01-31
If I click on ... 3 points and open the options, a click outside should just close it.


bitcoinuser:
Hi,

If I click on ...  3 points open the mini window, a click outside should just close it.
The way works now, when I click outside close the mini window and at same time open the thread.

https://user-images.githubusercontent.com/1178856/215719235-cd3dc81c-5c4a-4625-9e4b-7baddb6c42db.mp4

PR #304, opened 2023-02-14
Improve overflow menu button

joshr4:
In order to address https://github.com/v0l/snort/issues/177, I tried adding a background menu item which closes the sub-menu. I made this menu item cover a large area near the sub-menu. If a user clicks within this area, it will close the menu rather than performing the action underneath.

Rather than cover the whole page, I kept the size smaller. If a user scrolls away from the open sub-menu, they can still interact with other buttons and features.

I also increased the width of the three vertical dots menu button, so that its easier to click. Its current width is 4px, I increased this to 16px.

Here is a visual of the component, I temporarily set the color to blue so it is visible:

https://user-images.githubusercontent.com/14062229/218881434-0f46a7b9-2ac7-4241-b70e-2a0876497eb4.mov

### DIFF
--- a/packages/app/src/Element/NoteFooter.tsx
+++ b/packages/app/src/Element/NoteFooter.tsx
@@ -209,6 +209,13 @@ export default function NoteFooter(props: NoteFooterProps) {
   function menuItems() {
     return (
       <>
+        <div className="close-menu-container">
+          {/* This menu item serves as a "close menu" button;
+          it allows the user to click anywhere nearby the menu to close it. */}
+          <MenuItem>
+            <div className="close-menu" />
+          </MenuItem>
+        </div>
         {prefs.enableReactions && (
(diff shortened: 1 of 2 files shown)


packages/app/src/Element/NoteFooter.tsx
```diff
@@ -209,6 +209,13 @@ export default function NoteFooter(props: NoteFooterProps) {
   function menuItems() {
     return (
       <>
+        <div className="close-menu-container">
```
joshr4, line 212: adding a div wrapper here prevents the newly added menu item from altering the existing menu's styling/appearance.

v0l (approved):
Seems like a decent idea, thanks

v0l:
Please rebase, use `git rebase origin/main`

joshr4:
Yeah having some git troubles, needed to amend the older commits due to them being unverified

PR #305, opened 2023-02-15
fix: ran yarn prettier --w on repo

joshr4:
I noticed that the prettier check is failing on this readme PR https://github.com/v0l/snort/pull/300 (failing check https://github.com/v0l/snort/actions/runs/4178879144/jobs/7238346364).

I ran the prettier command on the repo and it fixed a few files.


h3y6e:
Hi.

`transifex.yml` should be ignored by [`.prettierignore`](https://github.com/v0l/snort/blob/main/packages/app/.prettierignore).

I think prettier and eslint should be installed as shared libraries (but doesn't seem to be within the scope of this PR).

v0l:
Fixed in #301

PR #314, opened 2023-02-15
feat: render kind 1 reposts

kphrx:


packages/app/src/Element/Note.tsx
```diff
@@ -65,6 +66,12 @@ export default function Note(props: NoteProps) {
   const { data, related, highlight, options: opt, ["data-ev"]: parsedEvent, ignoreModeration = false } = props;
   const [showReactions, setShowReactions] = useState(false);
   const ev = useMemo(() => parsedEvent ?? new NEvent(data), [data]);
+  if (
+    ev.Kind === EventKind.TextNote &&
+    ev.Tags.some((a, i) => a.Key === "e" && a.Marker === "mention" && ev.Content === `#[${i}]`)
+  ) {
+    return <NoteReaction data={data} key={data.id} />;
+  }
```
kphrx, line 74: really needs?

verbiricha: Don't think we need this, only place Note will be rendered outside of timeline is threads and I think is fine to render as a regular note with link to original there. We do the #[idx] substitution for content.

packages/app/src/Feed/TimelineFeed.ts
```diff
@@ -147,11 +147,21 @@ export default function useTimelineFeed(subject: TimelineSubject, options: Timel
...
         .map(a => a.tags.find(b => b[0] === "e"))
         .filter(a => a)
         .map(a => unwrap(a)[1]);
+      const repostsByKind1 = main.store.notes
+        .filter(
+          a =>
+            (a.kind === EventKind.Repost || a.kind === EventKind.TextNote) &&
+            a.tags.some(b => b[0] === "e" && b[3] === "mention")
+        )
+        .map(a => a.tags.find((b, i) => b[0] === "e" && b[3] === "mention" && a.content === `#[${i}]`))
```
verbiricha, line 161: Looks like we could extract this part of the logic to a `isTextRepost` function, seems to be used in a few places.

verbiricha (commented):
minor comments, other than that lgtm

packages/app/src/Feed/TimelineFeed.ts
kphrx: fixed https://github.com/v0l/snort/commits/615c2a974ab64a4d025ea4d4f382a2da631181f9

packages/app/src/Element/Note.tsx
kphrx: thanks, removed them in https://github.com/v0l/snort/commits/d911dc8da321409787efc0a7cff6530d37894ba1.

packages/app/src/Element/Note.tsx
```diff
@@ -98,7 +98,14 @@ export default function Note(props: NoteProps) {
   }, [reactions]);
   const positive = groupReactions[Reaction.Positive];
   const negative = groupReactions[Reaction.Negative];
-  const reposts = useMemo(() => dedupeByPubkey(getReactions(related, ev.Id, EventKind.Repost)), [related, ev]);
+  const reposts = useMemo(
+    () =>
+      dedupeByPubkey([
+        ...getReactions(related, ev.Id, EventKind.TextNote).filter(e => e.tags.some(tagFilterOfTextRepost(e, ev.Id))),
```
kphrx, line 104: missing `EventKind.TextNote` in `related`

verbiricha: I think is fine to only show kind1 reposts in timeline and not as part of note reactions, fetching text notes related to to every note can give us a bunch of results if is part of a thread and we are only interested in a subset (reposts) we can't filter by in the subscription. I think is fine we keep the related filter by kind and not show kind1 reposts here.

verbiricha:
Can you run `yarn format` on `packages/app`? Format check is failing. After that I'll merge it, thanks for your patience!

v0l (approved):

PR #315, opened 2023-02-15
`nostr` package part 1

ennmichael:
Refactoring the `nostr` package interface to be more idiomatic.

_Most changed files are renames, this is not a big PR._

This is the general layout I have in mind. `Event` is a union of interfaces discriminated on the `kind` field. The `Nostr` class is the client. The `onEvent` callback accepts a strongly-typed event as well as the raw event format for highly exotic uses. (IMO the only reason why the user might want the raw format is if there's a new NIP that isn't supported by our client yet and the user wants to support it.)

I still have to move a lot of code from the `Connection` type into the refactored code. Also I need to write the interfaces for all remaining events which is mostly manual labor.


packages/nostr/src/index.ts
```diff
@@ -1,87 +1,3 @@
-export * from "./System";
```
ennmichael, line 1: This was moved to `legacy/index.ts`, nothing was changed.

v0l (commented):
Seems like a good start, where is the connection code in `Nostr` its all todo still?

v0l (approved):
Ok i see we still have export legacy so its still using the legacy code

PR #316, opened 2023-02-15
Reduce space between the texts for selecting relays

h3y6e:
I think the spacing between `Read global from` and `[globalRelay]` is large and a bit misleading. Especially in Japanese, I thought.

after:
after

### DIFF
--- a/packages/app/src/Pages/Root.tsx
+++ b/packages/app/src/Pages/Root.tsx
@@ -88,22 +88,19 @@ export default function RootPage() {
     <>
       <div className="main-content">{pubKey && <Tabs tabs={tabs} tab={tab} setTab={setTab} />}</div>
       {isGlobal && globalRelays.length > 0 && (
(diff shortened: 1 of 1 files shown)


packages/app/src/Pages/Root.tsx
```diff
@@ -88,22 +88,19 @@ export default function RootPage() {
...
-          <div>
-            <select onChange={e => setRelay(e.target.value)}>
-              {globalRelays.map(a => (
-                <option key={a} value={a}>
-                  {new URL(a).host}
-                </option>
-              ))}
-            </select>
-          </div>
+        <div className="flex mb10 f-end">
```
verbiricha, line 91: can we put this div inside the previous `div.main-content`? this way it will have padding on mobile.

h3y6e: Ok, I fixed.

v0l (approved):

PR #323, opened 2023-02-17
German translations for snort

gandlafbtc:
Some translations, like `"{n,plural,=1{{n} new note} other{{n} new notes}}",` i didn't know to what extent they can be translated, so i left them unchanged


v0l:
:O You did it manually? Why not with https://crowdin.com/project/snort

gandlafbtc:
Oh! I didn't know about that! Should I do it again with the website?

gandlafbtc:
I know someone who would do Korean. Is there an option for that?

v0l:
> I know someone who would do Korean. Is there an option for that?

No its fine it should sync them again

> I know someone who would do Korean. Is there an option for that?

I will add Korean now! That would be awesome

v0l (approved):

PR #326, opened 2023-02-17
RequestBuilder / Core Refactor

v0l:
This PR has been hijacked by the `RequestBuilder` rewrite, the Gossip code is not used atm and will be hooked up to the new system in another PR.

Process goes like this:

1. Cache relay lists (ContactLists + 10002 Relay List) in `LoginFeed`
2. Use custom hook `useRelaysForFollows` which gives you `pickRelays` for a set of PubKeys
   1. Here we look at all pubkeys and count the number of pubkeys on each relay
   2. After this we pick the top 3 relays for each pubkey (relays are sorted by most popular first)
   3. Next we select all distinct relays in the list and get the pubkeys on each relay which we want to fetch
4. Expand `Subscriptions` by relay and pubkey set, for timeline subject `pubkey` we return a list of subs for each relay we want to query and by which pubkeys
5. Everything else works the same


verbiricha (approved):
tested ACK

v0l:
Needs more work

verbiricha:
I'm not getting any search results even with frequently used keywords such as "zap".

verbiricha:
Following this pubkey https://relay-lb.snort-social.pages.dev/p/npub1n0sturny6w9zn2wwexju3m6asu7zh7jnv2jt2kx6tlmfhs7thq0qnflahe but can't find any notes.

v0l:
Posting a reply doesn't show in thread, data is seen on feed though

PR #327, opened 2023-02-17
feat(relays): add export to clipboard and handle multiple relay urls

fernandolguevara:
image


packages/app/src/Pages/settings/Relays.tsx
```diff
@@ -31,6 +31,10 @@ const RelaySettingsPage = () => {
     }
   }
 
+  function exportRelays() {
+    navigator.clipboard.writeText(Object.keys(relays).join("\r\n"));
```
v0l, line 35: Maybe we can join them by one of the separators in the split below, why `\r\n`

packages/app/src/Pages/settings/Relays.tsx
```diff
@@ -41,29 +45,72 @@ const RelaySettingsPage = () => {
...
+        </div>
       </>
     );
   }
 
   function addNewRelay() {
-    if ((newRelay?.length ?? 0) > 0) {
-      const parsed = new URL(newRelay ?? "");
+    if (newRelay?.length) {
+      let seps = "; ,".split("");
```
v0l, line 68: Doesn't seem necessary to support multiple separators, space alone seems fine

packages/app/src/Pages/settings/Relays.tsx
fernandolguevara: I want the list to be splited by new lines, If I need to paste the it on x IM clients

packages/app/src/Pages/settings/Relays.tsx
fernandolguevara: I prefer to use this list of characters for delimited lists

v0l:
Rejected

fernandolguevara:
joya aborto de nutria 🤙🏻

PR #335, opened 2023-02-20
WIP: Add support for NIP-46

tiero:
This PR currently adds support to connect to Nostr Connect enabled remote signers.

https://www.loom.com/share/1523039fc92948019f4c4957e666cb2f

Not ready to merge as the `Connect` instance should run in the main top-level component so web abb can always catch the connect/disconnect event.

Also would be interesting to add the possibility to call the `sign_event` method, along the NIP-26 delegation.


v0l:

Issue #317, opened 2023-02-16
Relays listed on profile sometimes don't match relays listed in settings


SamSamskies:
**Describe the bug**
Earlier today I added wss://nostr-dev.wellorder.net/, but I later deleted it because it wasn't working. For some reason my profile on Snort is still displaying it. The relay page in the settings is correct though.

**To Reproduce**
Not sure how to reproduce. Try adding and then removing a relay.

**Expected behavior**
When I click the Relays pill on my profile, the listed relays match what's in my my settings.

**Screenshots**
Relays on settings page is correct
![Screenshot 2023-02-15 at 7 06 15 PM](https://user-images.githubusercontent.com/3655410/219232129-4e0aeca3-c482-4616-8fc8-27c2b5cd28d9.png)

Relays on profile is not correct
![Screenshot 2023-02-15 at 7 05 46 PM](https://user-images.githubusercontent.com/3655410/219232094-4911cd8c-aec8-4df6-885b-bfceb253318c.png)

verbiricha:
I think this is because we are caching the relay list result, we shouldn't so we always show the latest https://github.com/v0l/snort/blob/main/packages/app/src/Feed/RelaysFeed.tsx#L16

SamSamskies:
I did more research on this and when I query relays for kind 10002 for my account I get an outdated list of relays. @v0l @verbiricha are we forgetting to broadcast a kind 10002 event when updating the relays on Snort? Is that how [nip-65](https://github.com/nostr-protocol/nips/blob/524caa38563f49fc2aacbbe48b212fde0f24f97e/65.md) is supposed to work? If that's what we're supposed to do, we could use [Blastr (wss://nostr.mutinywallet.com)](https://github.com/MutinyWallet/blastr) to broadcast the 10002 update when a user updates their relays on Snort.

SamSamskies:
I think I see the bug in our code, I'll make a PR and we can discuss there if I'm misunderstanding how this is supposed to work.

verbiricha:
> are we forgetting to broadcast a kind 10002 event when updating the relays on Snort? Is that how [nip-65](https://github.com/nostr-protocol/nips/blob/524caa38563f49fc2aacbbe48b212fde0f24f97e/65.md) is supposed to work?

When saving relays settings we should be broadcasting to all the relays we are connected to and a few others but perhaps there is a bug there.

[...]

PR #337, opened 2023-02-20
Fix stale relays

SamSamskies:
resolves https://github.com/v0l/snort/issues/317

Please test with testing accounts.

### DIFF
--- a/packages/app/src/Element/RelaysMetadata.tsx
+++ b/packages/app/src/Element/RelaysMetadata.tsx
@@ -12,7 +12,9 @@ const RelayFavicon = ({ url }: { url: string }) => {
     .replace(/^ws:\/\//, "http://")
     .replace(/\/$/, "");
   const [faviconUrl, setFaviconUrl] = useState(`${cleanUrl}/favicon.ico`);
-  return <img className="favicon" src={faviconUrl} onError={() => setFaviconUrl(Nostrich)} />;
+  return (
+    <img className="favicon" src={faviconUrl} onError={() => setFaviconUrl(Nostrich)} alt={`favicon for ${url}`} />
+  );
 };
 
 interface RelaysMetadataProps {
--- a/packages/app/src/Element/RelaysMetadata.tsx
+++ b/packages/app/src/Element/RelaysMetadata.tsx
@@ -24,7 +24,7 @@ const RelaysMetadata = ({ relays }: RelaysMetadataProps) => {
     <div className="main-content">
       {relays?.map(({ url, settings }) => {
         return (
-          <div className="card relay-card">
+          <div key={url} className="card relay-card">
             <RelayFavicon url={url} />
             <code className="relay-url f-ellipsis">{url}</code>
             <div className="relay-settings">
--- a/packages/app/src/Feed/EventPublisher.ts
+++ b/packages/app/src/Feed/EventPublisher.ts
@@ -116,6 +116,15 @@ export default function useEventPublisher() {
         }
       }
     },
+    /**
+     * Write to every online relay using Blastr.
+     * https://github.com/MutinyWallet/blastr
+     */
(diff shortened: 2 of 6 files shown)


packages/app/src/Feed/EventPublisher.ts
```diff
@@ -251,18 +260,6 @@ export default function useEventPublisher() {
...
-        const ev = NEvent.ForPubKey(pubKey);
-        ev.Kind = EventKind.ContactList;
-        ev.Content = JSON.stringify(relays);
-        for (const pk of follows) {
-          ev.Tags.push(new Tag(["p", pk], ev.Tags.length));
-        }
-
-        return await signEvent(ev);
-      }
-    },
```
SamSamskies, line 265: I don't think we're supposed to do this. At least from reading [nip-02](https://github.com/nostr-protocol/nips/blob/master/02.md) and [nip-65](https://github.com/nostr-protocol/nips/blob/master/65.md) it doesn't seem like we're supposed to do this. Let me know if I'm misunderstanding something.

v0l: Yes we are supposed to do this, relays are stored in the contact list ofc!

SamSamskies: Doh. I couldn't find it in the docs. Where is this documented?

SamSamskies: Nip-02 says the value of `content` should be ignored.

v0l: Its not documented but this is actually where all clients store relays right now, NIP-65 is new

SamSamskies: Oh maybe that's why a lot of relays don't seem to support nip-65. Let me revert my changes and update to render the relays pulled from the data from the kind 3 event rather than kind 10002 event then. I think that should fix the stale relays bug.

SamSamskies:
I think the reason for the stale relays is the lack of support for nip-65. Switched to using the relay data from the contact list and that seems to fix the bug.

packages/app/src/Element/RelaysMetadata.tsx
```diff
@@ -12,7 +12,9 @@ const RelayFavicon = ({ url }: { url: string }) => {
     .replace(/^ws:\/\//, "http://")
     .replace(/\/$/, "");
   const [faviconUrl, setFaviconUrl] = useState(`${cleanUrl}/favicon.ico`);
-  return <img className="favicon" src={faviconUrl} onError={() => setFaviconUrl(Nostrich)} />;
+  return (
+    <img className="favicon" src={faviconUrl} onError={() => setFaviconUrl(Nostrich)} alt={`favicon for ${url}`} />
```
SamSamskies, line 16: This has nothing to do with the bug. Just fixed the lint warning..

packages/app/src/Element/RelaysMetadata.tsx
```diff
@@ -24,7 +26,7 @@ const RelaysMetadata = ({ relays }: RelaysMetadataProps) => {
     <div className="main-content">
       {relays?.map(({ url, settings }) => {
         return (
-          <div className="card relay-card">
+          <div key={url} className="card relay-card">
```
SamSamskies, line 29: Just fixing missing key console warning.

packages/app/src/Pages/settings/RelayInfo.tsx
```diff
@@ -76,6 +76,7 @@ const RelayInfo = () => {
             <div className="f-grow">
               {stats.info.supported_nips.map(a => (
                 <span
+                  key={a}
```
SamSamskies, line 79: Just fixing missing key console warning.

verbiricha (approved):

PR #341, opened 2023-02-21
feat: show nip-23 long form notes

verbiricha:
Initial support for nip-23 long form notes. note1l9l07ajtu2u8s7ej03k06a33gptqzef9j8utksf2mjpcf9hj8ntsqgk22e is a kind 1 note that replies to one of these.


packages/app/src/Element/Note.tsx
```diff
@@ -175,6 +176,16 @@ export default function Note(props: NoteProps) {
     }
   }, [inView, entry, extendable]);
 
+  function goToReplaceableEvent(e: React.MouseEvent, ev: NEvent) {
+    e.stopPropagation();
+    const d = ev.Tags.find(t => t.Key === "d")?.Original[1];
+    if (d) {
+      const href = encodeTLV(d, NostrPrefix.Address, [], ev.PubKey, ev.Kind);
+      window.open(`nostr:${href}`);
+      //window.open(`https://habla.news/a/${href}`);
```
verbiricha, line 185: not sure if any client support naddr atm, perhaps i should not make it clickable @v0l ?

SamSamskies: I think it's fine to use the habla.news link. You could allow user to specify choose a different client in user preferences if people complain about it.

SamSamskies:
I feel like the design for the long form note should be more obvious that it's a long form note and note a regular note. Any way to get some design input on this? Also, it's not very intuitive that you can click it to go see the full note somewhere else. I think it would be better to include a lot of the normal options such as share, copy ID, etc. Could just add a `View full note` option to the list which would take user to the note on habla.news.

verbiricha:
> I think it would be better to include a lot of the normal options such as share, copy ID, etc. Could just add a View full note option to the list which would take user to the note on habla.news.

Makes sense, will talk with Karnage about it, I like the idea of showing a preview by default and a "Read more" option to show the full article. I think we can show it in snort instead of linking elsewhere.

SamSamskies:
>  I think we can show it in snort instead of linking elsewhere.

Boom. Then you can check that nip-23 box in the README. 😂

PR #347, opened 2023-02-22
fix links in parentheses

SamSamskies:
resolves https://github.com/v0l/snort/issues/343

Screenshot 2023-02-21 at 6 59 42 PM

PR #350, opened 2023-02-24
Revert "Merge pull request #347 from v0l/fix-links-in-parentheses"

SamSamskies:
This regex seems to cause problems in some browsers. Reverting back to what it was before for now.


SamSamskies:
This fixes the problem of the site not loading in Safari and on mobile browsers.

v0l (approved):

PR #353, opened 2023-02-24
Small settings page stuff

w3irdrobot:
I noticed that the profile settings page was pretty unstyled compared to the preferences page. So I just updated it a little bit to make it match that page.

image


v0l (approved):
Good idea thanks!

PR #354, opened 2023-02-24
Change message unread color to purple

w3irdrobot:
This is more of a suggestion. Not a bug. I've noticed when looking at alot of DMs, though the unread background color is different, it's not different enough to make it easy to see what messages have unreads. So I updated the color to be the purple highlight color so it sticks out more. 

Before:

image

After:

image


v0l (approved):
Another good tweak, thanks!

PR #360, opened 2023-02-25
fix URL parsing edge cases

SamSamskies:
This should fix the following:
- URLs in parentheses
- URLs at the end of a sentence

Some links for testing:
URLs in parenthesis - https://fix-url-parsing-edge-cases.snort-social.pages.dev/e/note1phqc0hu77psunm49fdwjfdnwmmm9pzu9tq52wrkue7gceht9cfpqmajhs3

URLs at the end of a sentence - https://fix-url-parsing-edge-cases.snort-social.pages.dev/e/note1as9axphvvkhzagx62p5qgdh6jqq93d5f3pdueqpdwkf77y7ylqts8tuue2

### DIFF
--- a/packages/app/src/Util.test.ts
+++ b/packages/app/src/Util.test.ts
@@ -1,0 +1,40 @@
+import { splitByUrl } from "./Util";
+
+describe("splitByUrl", () => {
+  it("should split a string by URLs", () => {
+    const inputStr =
+      "@npub1q6mcr8t not https://example.com- sure what your stack is, https://example.com but I made a https://example.com! simple example (https://example.com) of how https://example.com/yo-yo https://example.example.com to do this https://example.com, https://example.com?q=asdf for Next.js apps hosted on Vercel https://example.com. Scarcity in money provides the incentive to create abundance in other things as there is a mechanism to reliably store value. https://i.imgur.com/rkqhjeq.png Every form of money that could be inflated by way of force or technological advancement has been.";
+    const expectedOutput = [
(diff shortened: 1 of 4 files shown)


packages/app/src/Util.test.ts
```diff
@@ -0,0 +1,40 @@
+import { splitByUrl } from "./Util";
+
+describe("splitByUrl", () => {
+  it("should split a string by URLs", () => {
+    const inputStr =
+      "@npub1q6mcr8t not https://example.com- sure what your stack is, https://example.com but I made a https://example.com! simple example (https://example.com) of how https://example.com/yo-yo https://example.example.com to do this https://example.com, https://example.com?q=asdf for Next.js apps hosted on Vercel https://example.com. Scarcity in money provides the incentive to create abundance in other things as there is a mechanism to reliably store value. https://i.imgur.com/rkqhjeq.png Every form of money that could be inflated by way of force or technological advancement has been.";
+    const expectedOutput = [
+      "@npub1q6mcr8t not ",
+      "https://example.com-",
```
SamSamskies, line 9: oops this isn't correct. the dash shouldn't match on a tld.

SamSamskies: maybe we don't need to worry about that case since i don't see a reason why anyone would type a dash directly after a URL without a space.

v0l: Yea that would just be an invalid TLD

v0l (approved):

PR #363, opened 2023-02-26
feat: add search page field autofocus

lujakob:
This PR adds the `autofocus` attribute to the search field input element on the search page. This improves the UX especially when a user is searching and adding multiple users in a row.


v0l (approved):
👍

PR #365, opened 2023-02-26
`nostr` package part 3

ennmichael:
Fill in the holes of missing functionality to the point where there's a simple test which passes :tada:. The test requires the `docker-compose` environment to be up and talks to the `nostr-rs-relay`. Change `onEvent`, `onNotice`, `onError` etc. into a single overloaded `on("event" | "notice" | "error", {})` method. Figure out a good `SignedEvent` abstraction, which puts together the event data, ID, and signature. Add eslint.

I also changed the callbacks to no longer be a list, because I thought it was more idiomatic. This is probably a mistake; instead, I should model all the event stuff based on Node's [EventEmitter](https://nodejs.org/api/events.html#class-eventemitter). These API changes will be done in one of the following PRs.


packages/nostr/src/client/conn.ts
```diff
@@ -137,25 +139,139 @@ export const enum OutgoingKind {
...
+  until?: number
+  limit?: number
+}
+
+async function parseIncomingMessage(data: string): Promise<IncomingMessage> {
+  const json = parseJson(data)
+  if (!(json instanceof Array)) {
+    throw new ProtocolError(`incoming message is not an array: ${data}`)
+  }
+  if (json.length === 3) {
```
v0l, line 181: This is very definitely going to break, we should not use the length because relays and clients implement NIP's in random orders and all NIP's are optional, only the tag from `json[0]` should be used to determine what sort of message this is

packages/nostr/src/client/conn.ts
```diff
@@ -137,25 +139,139 @@ export const enum OutgoingKind {
...
+}
+
+function serializeOutgoingMessage(msg: OutgoingMessage): string {
+  if (msg.kind === OutgoingKind.Event) {
+    return JSON.stringify(["EVENT", msg.signed.serialize()])
+  } else if (msg.kind === OutgoingKind.Subscription) {
+    return JSON.stringify([
+      "REQ",
+      msg.id.toString(),
+      serializeFilters(msg.filters),
```
v0l, line 229: Filters should be an array, you can have multiple `RawFilters` which get `OR`'d

packages/nostr/src/client/index.ts
```diff
@@ -18,62 +12,53 @@ export class Nostr {
...
-
+  on(on: "notice", cb: NoticeCallback | undefined | null): void
   /**
    * Add a new callback for errors.
    */
-  onError(cb: ErrorCallback): void {
-    this.#errorCallbacks.push(cb)
+  on(on: "error", cb: ErrorCallback | undefined | null): void
+  // TODO Also add on: ("subscribed", subscriptionId) which checks "OK"/"NOTICE" and makes a callback?
+  // TODO Also add on: ("sent", eventId) which checks "OK"/"NOTICE" and makes a callback?
```
v0l, line 39: Can be added later, we dont currently use OK callback on snort but it would be nice to get confirmation when trying to publish events

packages/nostr/src/client/index.ts
```diff
@@ -216,19 +202,39 @@ export class Nostr {
   /**
    * Publish an event.
    */
-  async publish(event: Event): Promise<void> {
+  async publish(event: Event, key: PrivateKey): Promise<void> {
```
v0l, line 205: We wont always have the private key so we need a way to publish events that are already signed

packages/nostr/src/event.ts
```diff
@@ -52,22 +51,160 @@ export interface UnknownEvent extends EventCommon {
...
   const event = {
     pubkey,
     createdAt,
   }
-  await checkSignature(raw, event)
   return (
-    createSetMetadataEvent(raw, event) ??
-    createTextNodeEvent(raw, event) ??
-    createUnknownEvent(raw, event)
+    parseSetMetadataEvent(raw, event) ??
```
v0l, line 201: Why not just switch on `raw.kind`

packages/nostr/src/event.ts
```diff
@@ -103,51 +240,57 @@ function createTextNodeEvent(
...
-  const hex = secp.utils.bytesToHex(bytes).toLowerCase()
-  if (hex.toString() !== raw.sig) {
-    throw new ProtocolError("invalid signature: ${hex}")
+function serializeEventContent(event: Event): string {
+  if (event.kind === EventKind.SetMetadata) {
+    // TODO Maybe rename userMetadata to content?
+    return JSON.stringify(event.userMetadata)
+  } else if (event.kind === EventKind.TextNote) {
+    // TODO Maybe rename note to content?
+    return event.note
```
v0l, line 261: Yes pls use `content` instead of `note`

packages/nostr/src/keypair.ts
```diff
@@ -25,14 +29,51 @@ export class PublicKey {
 export class PrivateKey {
   #hex: string
 
-  constructor(hex: string | Uint8Array) {
-    this.#hex = parseHex(hex)
+  /**
+   * Expects the key encoded as an nsec1-prefixed bech32 string, hex string, or byte buffer.
+   */
+  constructor(key: string | Uint8Array) {
+    this.#hex = parseKey(key, "nsec1")
```
v0l, line 36: The prefix is `nsec`, `1` is the bech32 seperator

v0l (approved):

PR #368, opened 2023-02-27
Feat/add spinner to button

lujakob:
Some information on the implementation:

* I decided to use css style visibility to hide the button text, to keep the width of the button on changing the state to the loading spinner

* I am actually used to handling the loading state outside the loading button, in the component where the button is used and the data handling of the button callback is handled. The loading state is then just passed down to the button. This way there is more control over the loading state, because it can be set and unset bevor and after the async request.

As it is right now, I think the loading is reset by a re-render of the button.

So I think in the long run, I think it makes sense to have the loading state handled in the parent component and pass it down to the button. I would need to refactor alls 4 places where the AsyncButton is used. What do you think?

* Would it make sense to rename AsyncButton to LoaderButton or SpinnerButton?

* Should I also apply AsyncButton to /settings/relays "Save" button?

* I added a "disabled" property to the AsyncButton, for future use. Maybe a button should be disalbed, as long as form fields are not valid.

### DIFF
--- a/packages/app/src/Element/AsyncButton.css
+++ b/packages/app/src/Element/AsyncButton.css
@@ -1,0 +1,27 @@
+button {
+  position: relative;
+}
+
+.spinner {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  top: 0;
+  left: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  animation-name: spin;
+  animation-duration: 800ms;
+  animation-iteration-count: infinite;
+  animation-timing-function: linear;
+}
+
+@keyframes spin {
+  from {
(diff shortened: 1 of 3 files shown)


v0l (changes requested):
I added a new Spinner for the fast zaps, can you use that one instead? Please rebase and you will see it

lujakob:
> I added a new Spinner for the fast zaps, can you use that one instead? Please rebase and you will see it

The spinner is very opaque thought, because of global css styles for disabled buttons having `opacity: 0.3` - this also affects the spinner. Please have a look, in case it might make sense to re-think buttons styles for the different states and maybe replace using `opacity` with changing color brightness for hover states.

PR #370, opened 2023-02-27
Fast Zaps ⚡ 

v0l:
Allows instant zapping with default zap amount

Long press zap icon to get old LNURL flow


verbiricha (approved):
⚡

packages/app/src/Element/NoteFooter.tsx
```diff
@@ -102,15 +118,39 @@ export default function NoteFooter(props: NoteFooterProps) {
...
+    if (zapping || e.isPropagationStopped()) return;
+
+    const lnurl = author?.lud16 || author?.lud06;
+    if (webln?.enabled && lnurl) {
+      setZapping(true);
+      try {
+        const handler = new LNURL(lnurl);
+        await handler.load();
+
+        const zap = await publisher.zap(prefs.defaultZapAmount * 1000, ev.PubKey, ev.Id);
```
v0l, line 131: Need to check `handler.canZap` should we allow Non-zap with fast button?

packages/app/src/LNURL.ts
```diff
@@ -0,0 +1,145 @@
...
+        throw new Error("Invalid LNURL: not a url");
+      }
+      this.#url = new URL(decoded);
+    } else if (lnurl.match(EmailRegex)) {
+      const [handle, domain] = lnurl.toLowerCase().split("@");
+      this.#url = new URL(`https://${domain}/.well-known/lnurlp/${handle}`);
+    } else if (lnurl.toLowerCase().startsWith("http")) {
+      this.#url = new URL(lnurl);
+    } else {
+      throw new Error("Invalid LNURL: could not determine service url");
```
v0l, line 24: Need to extract these messages

packages/app/src/Element/NoteFooter.tsx
verbiricha: I think falling back to non-zap if not supported is fine, still want the instant UX imo.

Issue #375, opened 2023-02-28
Display nostrnests room links as an iframe


SamSamskies:
**Is your feature request related to a problem? Please describe.**
![Screenshot 2023-02-28 at 1 21 04 PM](https://user-images.githubusercontent.com/3655410/221957129-e2c0ab85-eca3-4a93-8442-effa8e12ca83.png)

**Describe the solution you'd like**

Render room link as an iframe https://nostrnests.com/lyzsgu. The link should render in addition to the iframe so that users can open it in a new browser tab if they want.

```html
<iframe src="https://nostrnests.com/lyzsgu" allow="microphone *;" width="480" height="680"></iframe>
```

![Screenshot 2023-02-28 at 1 45 30 PM](https://user-images.githubusercontent.com/3655410/221962168-565047a7-9a30-4f0c-874c-3758779b9610.png)

PR #377, opened 2023-03-01
add Nostr Nests embed

SamSamskies:
resolves https://github.com/v0l/snort/issues/375

Not sure if it's a good idea to add this right now since Nosts Nests is still early in development and they don't actually have real embed code, but it could be a good way to get more people to check out the linked nests.

https://user-images.githubusercontent.com/3655410/222055058-31e371ce-d3bb-4a2d-9283-1c2cff0d5112.mp4

### DIFF
--- a/packages/app/src/Element/HyperText.tsx
+++ b/packages/app/src/Element/HyperText.tsx
@@ -24,7 +24,7 @@ import TidalEmbed from "Element/TidalEmbed";
 import { ProxyImg } from "Element/ProxyImg";
 import TwitchEmbed from "Element/TwitchEmbed";
 import AppleMusicEmbed from "Element/AppleMusicEmbed";
-import NostrNestsEmbed from "./NostrNestsEmbed";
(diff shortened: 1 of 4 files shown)


packages/app/src/Element/HyperText.tsx
```diff
@@ -23,6 +24,7 @@ import TidalEmbed from "Element/TidalEmbed";
 import { ProxyImg } from "Element/ProxyImg";
 import TwitchEmbed from "Element/TwitchEmbed";
 import AppleMusicEmbed from "Element/AppleMusicEmbed";
+import NostrNestsEmbed from "./NostrNestsEmbed";
```
v0l, line 27: `Element/NostrNestsEmbed`

v0l (commented):
Not sure if we should add this, might be a bad experience when its down all the time

@Semisol is this a good idea to add it now?

v0l (approved):

PR #378, opened 2023-03-01
Add people search

brugeman:
Trying to implement useful people search where results are ordered by 'popularity', not kind 0 update recency.

To enable the popularity sorting, I append " sort:popular" to the "search" filter - a way to pass options in NIP-50. I also have to disable loading of 'latest' events for this feed, bcs latest don't fit this sort order. And I had to use LIMIT_UNTIL and increase limit to 100 for this feed bcs scanning back with 'until' makes no sense with this sort order. And I had to disable the sorting by created_at, bcs of this sort order.

So it's a hack on a hack on a hack. Results look decent, but I don't like how it was achieved. 

Please take a look, maybe you'd have some suggestions.

### DIFF
--- a/packages/app/src/Feed/TimelineFeed.ts
+++ b/packages/app/src/Feed/TimelineFeed.ts
@@ -35,10 +35,8 @@ export default function useTimelineFeed(subject: TimelineSubject, options: Timel
 
     const sub = new Subscriptions();
     sub.Id = `timeline:${subject.type}:${subject.discriminator}`;
-    if (subject.type === "profile_keyword")
-      sub.Kinds = new Set([EventKind.SetMetadata]);
-    else
-      sub.Kinds = new Set([EventKind.TextNote, EventKind.Repost]);
+    if (subject.type === "profile_keyword") sub.Kinds = new Set([EventKind.SetMetadata]);
+    else sub.Kinds = new Set([EventKind.TextNote, EventKind.Repost]);
     switch (subject.type) {
       case "pubkey": {
         sub.Authors = new Set(subject.items);
--- a/packages/app/src/Feed/TimelineFeed.ts
+++ b/packages/app/src/Feed/TimelineFeed.ts
@@ -103,7 +103,7 @@ export default function useTimelineFeed(subject: TimelineSubject, options: Timel
       subLatest.Since = Math.floor(new Date().getTime() / 1000);
     }
     return subLatest;
-  }, [pref, createSub, subject.type]);
+  }, [pref, createSub]);
 
   const latest = useSubscription(subRealtime, {
     leaveOpen: true,
--- a/packages/app/src/Feed/TimelineFeed.ts
+++ b/packages/app/src/Feed/TimelineFeed.ts
@@ -91,7 +91,7 @@ export default function useTimelineFeed(subject: TimelineSubject, options: Timel
       }
     }
     return sub;
-  }, [until, since, options.method, pref, createSub, subject.type]);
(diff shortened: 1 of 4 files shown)


verbiricha:
> To enable the popularity sorting, I append " sort:popular" to the "search" filter - a way to pass options in NIP-50.

Really like how this looks/works but perhaps we could make this more evident by adding a "Sort by" selector with Recent/Popular and "Popular" preselected. Could look similar to global relay selector.

packages/app/src/Pages/SearchPage.tsx
```diff
@@ -1,22 +1,31 @@
...
   const [search, setSearch] = useState<string>();
   const [keyword, setKeyword] = useState<string | undefined>(params.keyword);
-  const allUsers = useQuery(keyword || "");
+  // tabs
+  const SearchTab = {
+    Posts: { text: formatMessage(messages.Posts), value: POSTS },
+    Profiles: { text: formatMessage(messages.People), value: PROFILES },
+  };
+  const [tab, setTab] = useState<Tab>(SearchTab.Posts);
+  const horizontalScroll = useHorizontalScroll();
```
verbiricha, line 28: Not needed here for just two tabs imo

brugeman: Agree, removed

brugeman:
Added the Sort dropdown on the people search tab.

packages/app/src/Feed/TimelineFeed.ts
```diff
@@ -49,8 +50,11 @@ export default function useTimelineFeed(subject: TimelineSubject, options: Timel
         sub.PTags = new Set(subject.items);
         break;
       }
-      case "keyword": {
-        sub.Kinds.add(EventKind.SetMetadata);
+      case "profile_keyword": {
+        sub.Search = subject.items[0];
```
v0l, line 54: You can set the kind in here instead then you dont need that if statement above

packages/app/src/Feed/TimelineFeed.ts
```diff
@@ -87,19 +91,19 @@ export default function useTimelineFeed(subject: TimelineSubject, options: Timel
       }
     }
     return sub;
-  }, [until, since, options.method, pref, createSub]);
+  }, [until, since, options.method, pref, createSub, subject.type]);
```
v0l, line 94: No need for subject.type here, createSub already has that dependency

packages/app/src/Feed/TimelineFeed.ts
brugeman: All subject.types except profile_keyword have TextNote & Repost kind, profile_keyword only has SetMetadata only. Seems like that if is necessary, or some version of it?

packages/app/src/Feed/TimelineFeed.ts
brugeman: Removed

packages/app/src/Feed/TimelineFeed.ts
v0l: Yea just overwrite the kinds in here and leave the old default kinds above without if statement?

```
sub.Kinds = new Set([a,b]);
...
case "profile_keyword": 
  sub.Kinds = new Set[c]);
  ...
```

brugeman: Ok, that's simpler, done.

v0l (approved):

verbiricha (approved):

v0l:

PR #382, opened 2023-03-02
Proposal: Remove SVGs from JSX

enjikaka:
Background: https://twitter.com/_developit/status/1382838799420514317

- Move "dumb" icons to icons.svg and use via new <Icon> element that renders an `<svg>` with `<use>`.
- Tidy up some alignments and DOM structure around icons (remove unneeded DOM nodes - better CSS formats)

Diffgifs:

compare

compare2


v0l (approved):
Nice one thanks!

PR #412, opened 2023-03-08
`nostr` package: vastly simplify the API

sistemd:
I've decided to simplify the API.

- Instead of classes, use type aliases for types which are really only strings (e.g. event ID, event signature, public/private key, hex strings). I'm only going to validate these when they enter the library functions. In the context of application development, having strong types is better, but for the purpose of keeping the API as simple as possible I decided it's best for the library to expose as few interfaces as possible, and leave the strong typing to the application.
- Don't separate the `RawEvent` and `Event` concepts. I had an important realization that the `RawEvent` and `Event` both carry the same data: `Event` represents a particular interpretation of that data. So instead of changing the fields of `Event` based on `kind`, the fields (i.e. the data) are always the same: those of `RawEvent`. Instead, depending on the `Event`'s `kind` field, the event will have additional _methods_ which allow the data to be interpreted in different ways. E.g. when `kind === EventKind.SetMetadata`, the event will have a method called `getUserMetadata` which parses the `content` as user metadata JSON. Similar for `EventKind.DirectMessage`, there will be a method `getMessage` which decrypts the message, etc. The TS type system is powerful enough to elegantly express this.
- Instead of separating the `SignedEvent` and `Event`, there is now a generic type `Unsigned<Event>` which just represents the event with the `sig`, `id`, and `pubkey` fields optionally omitted. Similar to the point above, the goal of this design is to help with unifying the event data. There's a `signEvent` method which fills in the missing fields. This also fits perfectly with NIP-07.
- Finally, I don't think that strong typing like this (i.e. "newtypes") are very idiomatic in the TS/JS ecosystem to begin with. I think that the ecosystem prefers simplicity.

Altogether I think this is way better than what I had before. I believe this is very close to what the final API will look like and hopefully there won't be any more big API refactors in the future, just implementing functionality from all remaining NIPs.


packages/nostr/src/event.ts
```diff
@@ -31,331 +32,255 @@ export enum EventKind {
...
+export async function parseEvent(event: RawEvent): Promise<Event> {
+  // TODO Validate all the fields. Lowercase hex fields, etc. Make sure everything is correct.
+  if (event.id !== (await calculateEventId(event))) {
+    throw new ProtocolError(
+      `invalid id ${event.id} for event ${JSON.stringify(
+        event
+      )}, expected ${await calculateEventId(event)}`
+    )
+  }
+  if (!schnorrVerify(event.sig, event.id, event.pubkey)) {
```
v0l, line 173: This is async, missing await

v0l (commented):
I was happy enough with the strong types, but maybe it was a bit overkill

One thing that we should try to accomodate is allowing people to do unsupported NIP's dont want a lib that needs to be constantly updated

One place this is obviously a problem in its current form is the REQ filters, those have a bunch of other values which are not implemented in this lib

v0l (approved):

PR #416, opened 2023-03-08
Add Wavlake embed

blastshielddown:
This enables the display of an embedded music player when people share links to tracks on Wavlake. The embedded player also is capable of boosting the track if the Alby extension (WebLN) is available and the wallet is capable of sending keysends. The boost button is not visible if WebLN is not available.

Screenshot 2023-03-08 at 5 31 08 PM

### DIFF
--- a/packages/app/public/index.html
+++ b/packages/app/public/index.html
@@ -8,7 +8,7 @@
     <meta name="description" content="Fast nostr web ui" />
     <meta
       http-equiv="Content-Security-Policy"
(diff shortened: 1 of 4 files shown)


SamSamskies (changes requested):
lgtm, but can you run `yarn workspace @snort/app format` to prettify your changes?

also, what do you think about removing the max width of the player? imo it looks better and that's how all the other player embeds are.

Screenshot 2023-03-08 at 5 50 21 PM

blastshielddown:
Thanks @SamSamskies Prettified changes as requested.

Also yes, good call on layout. I extended the max width of the embedded player to take up the width of the component. See screenshot below.

As for the Alby failure, what you're seeing is expected if the wallet/node you're connected to does not support keysends. Planning to support more widely available payment methods in the future but keysend is available now and would like to roll this out for anyone who's interested in trying it.

Screenshot 2023-03-08 at 6 31 55 PM

SamSamskies (approved):
🚀

SamSamskies:
@blastshielddown looks like you need to sign your commits. could you rebase and sign your commits?

blastshielddown:
@SamSamskies Rebased and signed

PR #424, opened 2023-03-10
Added key attr to Tabs to remove React warning

w3irdrobot:
While working on something else, I noticed a warning in the console about the Tabs not having unique `key` attributes. This just adds that to shut React up. I doubt any performance increases actually come from this. :laughing:

### DIFF
--- a/packages/app/src/Element/Tabs.tsx
+++ b/packages/app/src/Element/Tabs.tsx
@@ -32,7 +32,7 @@ const Tabs = ({ tabs, tab, setTab }: TabsProps) => {
   return (
     <div className="tabs" ref={horizontalScroll}>
       {tabs.map(t => (
-        <TabElement tab={tab} setTab={setTab} t={t} />
+        <TabElement key={t.value} tab={tab} setTab={setTab} t={t} />
       ))}
     </div>
   );


packages/app/src/Element/Tabs.tsx
```diff
@@ -32,7 +32,7 @@ const Tabs = ({ tabs, tab, setTab }: TabsProps) => {
   return (
     <div className="tabs" ref={horizontalScroll}>
       {tabs.map(t => (
-        <TabElement tab={tab} setTab={setTab} t={t} />
+        <TabElement key={t.text} tab={tab} setTab={setTab} t={t} />
```
v0l, line 35: There is `t.value` also?

v0l: `value` is the index of the tab so i suppose its actually meant to be unique unlike text

w3irdrobot: yeah but there's nothing in here that i can see that requires the value to be unique. although looking at the implementation of `TabElement`, it appears that is the assumption. otherwise, multiple tabs would be "active." i'll update the PR to use `value`.

v0l (approved):

PR #425, opened 2023-03-10
NIP06 support

w3irdrobot:
This PR updates the new private key generation to be [NIP06-compliant](https://github.com/nostr-protocol/nips/blob/master/06.md). This basically means the entropy is generated using a mnemonic phrase per BIP39 and used as the seed for a BIP32 master key. The private key for the account is then derived from mater key using the designated derivation path of `m/44'/1237'/0'/0/0`. Support for restoring an account from a mnemonic key is also supported. Note: restoring from a phrase assumes the account was created using a phrase at the beginning.


packages/app/src/State/Login.ts
```diff
@@ -253,6 +258,11 @@ export interface SetFollowsPayload {
   createdAt: number;
 }
 
+export interface SetGeneratedKeyPayload {
+  key: HexKey;
+  entropy: HexKey;
```
v0l, line 263: only a `string` for entropy right?

packages/app/src/State/Login.ts
```diff
@@ -315,12 +325,16 @@ const LoginSlice = createSlice({
...
-    setGeneratedPrivateKey: (state, action: PayloadAction<HexKey>) => {
+    setGeneratedPrivateKey: (state, action: PayloadAction<SetGeneratedKeyPayload>) => {
       state.loggedOut = false;
       state.newUserKey = true;
-      state.privateKey = action.payload;
-      window.localStorage.setItem(PrivateKeyItem, action.payload);
-      state.publicKey = secp.utils.bytesToHex(secp.schnorr.getPublicKey(action.payload));
+      state.privateKey = action.payload.key;
+      state.generatedEntropy = action.payload.entropy;
+      window.localStorage.setItem(PrivateKeyItem, action.payload.key);
```
v0l, line 333: If you don't store the entropy in local storage then on-reload you will lose the mnemonic, also its a bit confusing that in some places its called mnemonic and here it's called entropy.

Isn't the mnemonic just a 256bit value the same as a bare private key?

w3irdrobot: wouldn't the point be to lose the mnemonic on reload? usually mnemonics are not kept around for security reasons. i tried to use the words mnemonic and entropy in different places to signal what is being used. the mnemonic phrase and the entropy in hex or uint8array can be derived from each other but are not used in the same places for the same reason. for example, i don't pass the mnemonic phrase into `fromMasterSeed`. i pass the entropy bytes derived from the mnemonic phrase into `fromMasterSeed`. however, i can make it the same if that's preferred.

packages/app/src/State/Login.ts
w3irdrobot: updated

packages/app/src/State/Login.ts
v0l: No sorry you're right, i see they are actually different things, i thought you were storing the mnemonic always.

And in regards to not keeping the value, you done it perfectly, i see its cleared when exiting the discover flow

v0l (approved):

PR #434, opened 2023-03-12
`nostr` package: more tests

sistemd:
Add a better test framework allowing the relay to be restarted before each test. Add tests for the DM functionality. Add a `README` file with a progress tracker.


packages/nostr/src/crypto.ts
```diff
@@ -104,7 +104,7 @@ export async function aesEncryptBase64(
   plaintext: string
 ): Promise<AesEncryptedBase64> {
   const sharedPoint = secp.getSharedSecret(sender, "02" + recipient)
-  const sharedKey = sharedPoint.slice(2, 33)
+  const sharedKey = sharedPoint.slice(1, 33)
```
v0l, line 107: why slice 1? isnt this hex?

v0l: How did this work if its been wrong the whole time??

PR #435, opened 2023-03-13
fix: private key strength should be 256 bits.

susumuota:
Added signature.

---

Hi! Thank you for creating a great client!

I think I found an issue about NIP-06 PR #425.

`bip39.generateMnemonic(wordlist)` generates 128 bit strength private keys.
In my opinion, it should be 256 bit if you replace `secp.utils.randomPrivateKey()` which generates 32 byte (256bit) private keys.


v0l (approved):

v0l:
cd89da39ddf27a75cc7cb0023ec61ed62749096b

PR #439, opened 2023-03-14
Update Wavlake embed URL, add support for album & artist links

blastshielddown:
As described in the title, I've created a new subdomain for embeds from Wavlake to simplify the URL parsing. This update also adds support for links to artist and album pages (in addition to tracks).


packages/app/src/Const.ts
```diff
@@ -168,4 +168,4 @@ export const MagnetRegex = /(magnet:[\S]+)/i;
 /**
  * Wavlake embed regex
  */
-export const WavlakeRegex = /player\.wavlake\.com\/(track)\/([.a-zA-Z0-9-]+)/;
+export const WavlakeRegex = /player\.wavlake\.com\/([.a-zA-Z0-9-/]+)/i;
```
SamSamskies, line 171: What do you think about making this regex more specific? This is pretty general and even matches on the RSS feed links e.g. https://player.wavlake.com/feed/e4c5ff61-ef60-40e9-a289-82b1027824ae. If anyone ever shared their RSS feed link, it would display an iframe with a 404 instead of their RSS feed link.

packages/app/src/Element/WavlakeEmbed.tsx
```diff
@@ -1,14 +1,14 @@
...
   return (
     <iframe
       style={{ borderRadius: 12 }}
       src={convertedUrl}
       width="100%"
-      height="360"
+      height="400"
       frameBorder="0"
-      allow="autoplay; clipboard-write; encrypted-media; fullscreen; picture-in-picture"
+      allow=""
```
SamSamskies, line 11: I believe this allows all permissions. I think it would be better to allow specific permissions you need or remove it if no permissions are needed.

```suggestion
```

SamSamskies (changes requested):

blastshielddown:
Updated with suggested changes @SamSamskies

SamSamskies (approved):
🚀

PR #459, opened 2023-03-24
add setting for rewriting twitter links to nitter

w3irdrobot:
something i thought would be nice would be to automate the rewriting of twitter links to point to nitter instead for those who would prefer this. this adds a setting that is off by default. if on, it will do a simple rewrite of the twitter link to nostr by updating the hostname.


packages/app/src/Element/HyperText.tsx
```diff
@@ -110,12 +110,25 @@ export default function HyperText({ link, creator }: { link: string; creator: He
...
         }
-      } else if (tweetId) {
+      } else if (tweetId && !pref.rewriteTwitterPosts) {
         return (
           <div className="tweet" key={tweetId}>
             <TwitterTweetEmbed tweetId={tweetId} />
           </div>
         );
+      } else if (pref.rewriteTwitterPosts && url.hostname == "twitter.com") {
+        url.host = "nitter.at";
```
v0l, line 120: Isnt the host nitter.net ?

w3irdrobot: ahh so it looks like `nitter.at` is one of [the other known public instances](https://github.com/zedeus/nitter/wiki/Instances#public), but it's not the official one. i'll update the host to the official one.

v0l (approved):

PR #461, opened 2023-03-25
Nostr links

v0l:
Use `nevent` / `nprofile` links when relay data is known

- Text notes ALWAYS have relay information so share links and clicking on threads will now always use `nevent` 
- When relay hints exist for `e` tags then `nevent` links will be created (mentions or `re:` tags in note header)
- `nostr:` or `web+nostr:` links will be parsed and converted into clickable links within snort (except `naddr`)
- Profile QR now uses `nprofile` as well as offering copy link

## Todo

- [ ] Relays data is currently not used yet when opening `nevent` / `nprofile`, i will add this in another PR when the core refactor is done (ephemeral connections support)

### DIFF
--- a/packages/app/src/Element/Copy.tsx
+++ b/packages/app/src/Element/Copy.tsx
@@ -5,14 +5,15 @@ import { useCopy } from "useCopy";
 export interface CopyProps {
   text: string;
   maxSize?: number;
+  className?: string;
 }
-export default function Copy({ text, maxSize = 32 }: CopyProps) {
+export default function Copy({ text, maxSize = 32, className }: CopyProps) {
   const { copy, copied } = useCopy();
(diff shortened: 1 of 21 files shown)


v0l:
Test post: https://nostr-links.snort-social.pages.dev/e/nevent1qqsy08urw6ztcfnktmw6gq67su4p6h2833zy9n33wzkkn4vhqhd2h7cpz9mhxue69uhkummnw3ezuamfdejj7eh0q6l

verbiricha:
We probably should add some margin under user in QR modal

Screenshot 2023-03-26 at 19 16 10

verbiricha:
Tested ACK ✨

verbiricha (approved):

PR #488, opened 2023-04-05
NIP-94 file headers

v0l:
Supporting NIP-94 file headers for void.cat uploader

Ref: https://github.com/nostr-protocol/nips/pull/337

Test post: https://nip94.snort-social.pages.dev/e/note1mns2hkjns8dtr724a7qfj02aw0vle6953lqh5f3zaexrztlszreqd4vmye


frbitten:
thanks for implementing NIP-94

v0l:
> thanks for implementing NIP-94
> 
> Could have a global in snort just to list in shared NIP-94. As a file share index

Could do, but maybe it's better left to microapps, trying to keep Snort "social" like

There are already so many features that it's getting hard to maintain

PR #490, opened 2023-04-05
`nostr` package: get tests passing in the browser

sistemd:
I can now truly call it an isomorphic library: it does in fact work in the browser as well as node. Will implement NIP-07 next.


v0l:
`Module not found: Error: Can't resolve '@snort/nostr' in '/opt/buildhome/repo/packages/app/src'`

v0l (approved):

PR #492, opened 2023-04-06
fix #477

### LINKED ISSUE #477: Datas not loading or Loading Slow
**Describe the bug**
This issue describes an issue i have been facing for some months now, it's not loading up data, and sometimes loading it up slowly. 

**To Reproduce**

* Check on any profile 

**Expected behavior**
Data should be loaded immediately i try fetching the data 

**Screenshots**
Screenshot from 2023-04-01 04-31-27

vivganes:
This is a proposal for fixing the issue mentioned in #477 

Idea is to be transparent to the user about what the client is doing under the hood (searching from different time ranges) so that the user can take corrective action themselves.

image


v0l:
I thought about this solution too but im not sure if i like it

vivganes:
> I thought about this solution too but im not sure if i like it

Would it be better if we show text like "searching for stuff from x days/months ago" instead of the exact dates?

PR #498, opened 2023-04-07
fix #496

### LINKED ISSUE #496: Bookmarks seem to be broken
**Describe the bug**
My bookmarks are not rendering anymore.

Screenshot 2023-04-06 at 12 06 39 PM

**To Reproduce**

1. Go to my profile https://snort.social/p/npub1vp8fdcyejd4pqjyrjk9sgz68vuhq7pyvnzk8j0ehlljvwgp8n6eqsrnpsw
2. Click on Bookmarks

vivganes:
Screenshot after fixing

image

### DIFF
--- a/packages/app/src/Element/Bookmarks.tsx
+++ b/packages/app/src/Element/Bookmarks.tsx
@@ -45,7 +45,11 @@ const Bookmarks = ({ pubkey, bookmarks, related }: BookmarksProps) => {
         .filter(b => (onlyPubkey === "all" ? true : b.pubkey === onlyPubkey))
         .map(n => {
           switch (n.kind) {
-            case EventKind.TextNote:
+            case EventKind.Reaction:
+            case EventKind.Repost:
+            case EventKind.ZapReceipt:
+              return <NoteReaction data={n} key={n.id} />;
+            default:
               return (
                 <Note
                   key={n.id}
@@ -54,19 +58,7 @@ const Bookmarks = ({ pubkey, bookmarks, related }: BookmarksProps) => {
                   options={{ showTime: false, showBookmarked: true, canUnbookmark: loginPubKey === pubkey }}
                 />
               );
-            case EventKind.Reaction:
-            case EventKind.Repost:
-            case EventKind.ZapReceipt:
-              return <NoteReaction data={n} key={n.id} />;
           }
-          return (
-            <Note
-              key={n.id}
-              data={n}
-              related={related}
-              options={{ showTime: false, showBookmarked: true, canUnbookmark: loginPubKey === pubkey }}
-            />
-          );
         })}
     </div>
   );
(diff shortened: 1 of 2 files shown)


packages/app/src/Element/Bookmarks.tsx
```diff
@@ -43,14 +44,21 @@ const Bookmarks = ({ pubkey, bookmarks, related }: BookmarksProps) => {
...
-              data={n}
-              related={related}
-              options={{ showTime: false, showBookmarked: true, canUnbookmark: loginPubKey === pubkey }}
-            />
-          );
+          switch (n.kind) {
+            case EventKind.Reaction:
+            case EventKind.Repost:
+            case EventKind.ZapReceipt:
+              return <NoteReaction data={n} key={n.id} />;
```
v0l, line 51: I dont think this supports removing the bookmark or showing that its a bookmark like we have with the `Note` kind, would need to support these actions too

vivganes: Got it! Let me work on it

packages/app/src/Element/Bookmarks.tsx
```diff
@@ -19,7 +18,7 @@ const Bookmarks = ({ pubkey, bookmarks, related }: BookmarksProps) => {
   const [onlyPubkey, setOnlyPubkey] = useState<HexKey | "all">("all");
   const loginPubKey = useSelector((s: RootState) => s.login.publicKey);
   const ps = useMemo(() => {
-    return [...new Set(bookmarks.map(ev => ev.pubkey))];
+    return [...new Set(bookmarks.filter(ev => ev.kind === EventKind.TextNote).map(ev => ev.pubkey))];
```
v0l, line 21: Maybe we should look upstream as to why we have these events in here? This could be a bug from the refactor, check in the `useBookmarksFeed` or one of those feeds for bookmarks to see why we're getting likes

vivganes: On it now

vivganes: we seem to be filtering the Pinned feed also in the same way as I did

https://github.com/v0l/snort/blob/ecee50caa94e55b1e4dc9b666aa719290547a1ff/packages/app/src/Pages/ProfilePage.tsx#L229

From my research, it looks like we need the non TextNote events to display the number of likes, zap, etc on the card

v0l: Yea seems like we should split the lists here: https://github.com/v0l/snort/blob/ecee50caa94e55b1e4dc9b666aa719290547a1ff/packages/app/src/Pages/ProfilePage.tsx#L94

`related` is always for reactions etc so the filtering can be added here, or change the `Bookmarks` element so that it only accepts `bookmarks` and not `related` and continue to do the filtering inside

vivganes: Done. Moved the filtering to `ProfilePage.tsx`

v0l (approved):

PR #499, opened 2023-04-07
fix #495

### LINKED ISSUE #495: Login form needs to protect privacy of the secret key
**Is your feature request related to a problem? Please describe.**
The current login form uses an ordinary unmasked text field to input the key. This doesn't protect the privacy of the user as attacker could take a picture of the screen and use it to login later.

**Describe the solution you'd like**
Use a password input field instead, optionally allowing the user to reveal the text input on demand.

**Describe alternatives you've considered**
Nothing.

**Additional context**
Additional advantage is that password managers like bitwarden can be used to auto-fill without much hassle.

vivganes:


packages/app/src/Pages/Login.tsx
```diff
@@ -279,7 +279,7 @@ export default function LoginPage() {
           <div className="flex">
             <input
               dir="auto"
-              type="text"
+              type="password"
```
v0l, line 282: This doesnt really make sense when you want to login with the NIP-5 handle or the pubkey, maybe this should switch to password type when it starts with nsec

vivganes: Done!

packages/app/src/Pages/Login.tsx
```diff
@@ -279,12 +279,20 @@ export default function LoginPage() {
           <div className="flex">
             <input
               dir="auto"
-              type="text"
+              type={key?.startsWith("nsec") ? "password" : "text"}
```
v0l, line 282: We also allow logging in with NIP-06 menemic, that should be masked too, but also makes it hard to enter if you can just typing the words.

It might be difficult to implement this properly with a single input.

packages/app/src/Pages/Login.tsx
```diff
@@ -279,12 +279,20 @@ export default function LoginPage() {
...
               placeholder={formatMessage(messages.KeyPlaceholder)}
               className="f-grow"
               onChange={e => setKey(e.target.value)}
             />
           </div>
+          <p className="login-note">
+            {key?.startsWith("nsec") && (
+              <span className="highlight">
+                We noticed that you are entering a secret key starting with nsec. We are masking your input to protect
```
v0l, line 291: All strings should be wrapped in the `FormattedMessage` element so that we can support translations

packages/app/src/Pages/Login.tsx
vivganes: To properly implement privacy for nsec and mnemonics, I did the following things:
1. Display an 'eye' icon next to the input field to reveal or hide the input, when the user clicks it
2. When the user enters a nsec or mnemonic, show a suggestion to hide the input using the 'eye' icon (if it is not hidden already)

packages/app/public/icons.svg
```diff
@@ -354,5 +354,19 @@
         stroke-linejoin="round"
       />
     </symbol>
+    <symbol id="openeye" viewBox="0 0 24 24" fill="none">
```
v0l, line 357: Can you use the eye icons from Figma? https://www.figma.com/file/Q79i1WVA1bIAzYaLihgteN/Snort?node-id=68-4951&t=DDTOEgw4924vgYGt-0

packages/app/src/Pages/Login.tsx
```diff
@@ -279,12 +294,28 @@ export default function LoginPage() {
...
+              size={30}
+              className="highlight btn-sm pointer"
+              onClick={flipMask}
+            />
           </div>
+          <p className="login-note">
+            {suggestMasking && !isMasking && (
+              <span className="highlight">
+                <FormattedMessage
+                  defaultMessage="Please mask your input by clicking the eye icon to protect your privacy"
```
v0l, line 313: I think people know what the eye is for, maybe this message is not needed

packages/app/public/icons.svg
vivganes: @v0l I dont have access to download icons from here..  Can you help me with the icons please? Or, any way to download is also helpful.

v0l: If you just sign into Figma it will let you, you just need any account

packages/app/src/Pages/Login.tsx
vivganes: removed this message now

packages/app/public/icons.svg
vivganes: Changed the icons to the ones from Figma

packages/app/src/Pages/Login.tsx
```diff
@@ -279,11 +285,17 @@ export default function LoginPage() {
...
+              type={isMasking ? "password" : "text"}
               placeholder={formatMessage(messages.KeyPlaceholder)}
               className="f-grow"
               onChange={e => setKey(e.target.value)}
             />
+            <Icon
+              name={isMasking ? "openeye" : "closedeye"}
+              size={30}
+              className="highlight btn-sm pointer"
+              onClick={flipMask}
```
v0l, line 297: `onClick={() => setMasking(s => !s)}`

v0l (approved):

PR #510, opened 2023-04-12
add ability to paste image from clipboard

### LINKED ISSUE #502: Allow pasting images from clipboard
**Is your feature request related to a problem? Please describe.**
When I want to upload a screenshot to Snort, I first need to store it in my local desktop. That's a wee bit inconvenient when compared to flows in Telegram, etc.

**Describe the solution you'd like**
 I want to be able to paste from my clipboard.

**Describe alternatives you've considered**
None

**Additional context**
I am willing work on this feature
Probable solution: https://stackoverflow.com/a/73659331

vivganes:
Fixes #502


packages/app/src/Element/NoteCreator.tsx
```diff
@@ -205,14 +223,36 @@ export function NoteCreator() {
...
+        return /^image\//.test(item.type);
+      });
+      if (items.length === 0) {
+        return;
+      }
+
+      const item = items[0];
+      const blob = item.getAsFile();
+      if (blob) {
+        const file = new File([blob], "filename.jpg", { type: "image/jpeg", lastModified: new Date().getTime() });
```
v0l, line 240: You dont need to do this, `uploadFile` should accept `Blob` too

PR #513, opened 2023-04-14
fix text layout

mattn:
before

image

after

image

### DIFF
--- a/packages/app/src/Element/Text.css
+++ b/packages/app/src/Element/Text.css
@@ -3,6 +3,8 @@
   line-height: 24px;
   white-space: pre-wrap;
   word-break: break-word;
+  margin-top: 12px;
+  margin-bottom: 12px;
 }
 
 .text a {


packages/app/src/Element/Text.css
```diff
@@ -3,6 +3,8 @@
   line-height: 24px;
   white-space: pre-wrap;
   word-break: break-word;
+  margin-top: 12px;
```
v0l, line 6: Not sure if this is the best solution, where does the extra padding come from, can we just make sure to use `overflow: hidden` on those elements because this will just add padding thats not needed for non-markdown

mattn: Another issue.

image

I can't read this Chinese text (because I'm Japanese), but this is difficult to read for them too.

v0l: We probably need to adjust the heading element line-height here, i think it was raised before as an issue

v0l:

PR #521, opened 2023-04-17
`nostr` package: delegation

sistemd:
Add delegation. Refactor the event types into classes with an `EventCommon` base class that contains some common methods. Get rid of `RawEvent`. The `EventCommon` type should not be exposed to the user - hopefully this is doable. If this is not doable, then I will refactor it away. Its only purpose is to allow some code to be "easily reused" - it's not necessary for type correctness and it has no utility beyond reusing code.

There are two more things to do to make these classes more ergonomic:

- Improve the factory methods (e.g. `TextNote.create`, `SetMetadata.create`). All parameters should be optional, and the ones that are common to all factory methods can be moved into their own type and reused in every factory method.
- Improve the event constructors. For example, the `userMetadata` getter of the `SetMetadata` event does some validation and throws if `this.content` is invalid. This pattern repeats in all events. These kinds of validation should instead be done in the event constructor.

Also as finishing touches, much of the docs of this lib will need to be vastly improved in the future. Not sure how close I am to that :-)


v0l:
```


21:11:30.294 | TS18048: 'window.nostr' is possibly 'undefined'.
-- | --
21:11:30.294 | 188 \|     return await EventExt.decryptData(raw.content, privKey, pubKey);
21:11:30.294 | 189 \|   } else {
21:11:30.295 | > 190 \|     return await barrierNip07(() => window.nostr.nip04.decrypt(pubKey, raw.content));
21:11:30.295 | \|                                     ^^^^^^^^^^^^
21:11:30.295 | 191 \|   }
21:11:30.295 | 192 \| }
21:11:30.295 | 193 \|


```

v0l:

Issue #501, opened 2023-04-08
nostr:npub text is displayed as `@` and is a broken link


SamSamskies:
**Describe the bug**
In this note, I wrote the literal text `nostr:npub`, but it's getting displayed as `@` and is a broken link.

<img width="610" alt="Screenshot 2023-04-08 at 12 20 25 PM" src="https://user-images.githubusercontent.com/3655410/230736993-448dbe64-1d44-4d53-a512-ef0eab078dad.png">

**To Reproduce**

1. https://snort.social/e/nevent1qqsdham43eqwnkgujpkz0epmpwhh7h9n8ruasvvdv4acvlyvtutvduq3z4tjk

**Expected behavior**
`nostr:npub` should display the literal text `nostr:npub` and not be replaced with a link.

PR #528, opened 2023-04-19
don't render nostr:npub as a link

SamSamskies:
resolves https://github.com/v0l/snort/issues/501

People often type the literal text `nostr:npub` when talking about nostr URIs. That shouldn't be converted to a link.

### DIFF
--- a/packages/app/src/Util.ts
+++ b/packages/app/src/Util.ts
@@ -499,7 +499,11 @@ export function validateNostrLink(link: string): boolean {
       return false;
     }
 
-    return parsedLink.id.length === 64;
+    if (parsedLink.type === NostrPrefix.PublicKey || parsedLink.type === NostrPrefix.Note) {
+      return parsedLink.id.length === 64;
+    }
+
+    return true;
   } catch {
     return false;
   }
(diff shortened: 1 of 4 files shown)


v0l (changes requested):
There is also an issue where parsing an invalid TLV will break the note, for example trying to preview note with content `nostr:nevent1xxx` 

Fixing this should also fix the invalid `nostr:npub`

SamSamskies:
I looked into it and that seems to be a separate issue, but I can try and fix both with this PR.

packages/app/src/Util.ts
```diff
@@ -491,6 +491,20 @@ export interface NostrLink {
...
 
+export function validateNostrLink(link: string): boolean {
+  try {
+    const parsedLink = parseNostrLink(link);
+
+    if (!parsedLink) {
+      return false;
+    }
+
+    return parsedLink.id.length === 64;
```
SamSamskies, line 502: Any valid ID should be 64 chars, right?

SamSamskies:
I think I was able to fix both issues.

packages/app/src/Util.ts
v0l: No, the ID type for naddr, is the d-tag, maybe should tadd some tests to this?

SamSamskies:
@v0l fixed the logic to work for all nip-19 types and added unit tests.

v0l (approved):

PR #529, opened 2023-04-19
Update regex for Wavlake embeds, ignore zine subdomain

blastshielddown:
This fixes an issue where shares from our new zine subdomain are getting blanked out because the regex is incorrectly converting them to embed links.


SamSamskies (approved):
🚀

PR #536, opened 2023-04-20
feat: broadcast or rebroadcast note to specific relays

DanConwayDev:
The rationale for this feature is that users should have greater control over which relays they publish to on a case by case basis. They should also be able to change their mind (hence the rebroadcast feature).

Implementation: This adds two related features to different sections of the UI:

### 1. broadcast note to specific relays

Added a new section to the advanced options in NoteCreator which is used for both new notes and replies. This lists the users write relays and enables users to uncheck relays they do not wish to broadcast the note to.
When one or more relays are unchecked the submission is diverted from ```publisher.broadcast``` to ```publisher.broadcastAll``` with the desired relays specified.

image

### 2. rebroadcast user's own note to specific relays

Replaces 'block' from the 'more' menu on a given note with 'Broadcast Again' if the note was authored by the logged in user. The logic here is that the user won't want to block themselves. We also don't think they should be broadcasting other people's notes to new relays. When the new option is selected a modal opens where the user can uncheck the relays they do not want to (re)broadcast to. 

image

### DIFF
--- a/packages/app/src/Element/NoteCreator.tsx
+++ b/packages/app/src/Element/NoteCreator.tsx
@@ -18,6 +18,7 @@ import {
   setActive,
   setPreview,
   setShowAdvanced,
+  setSelectedCustomRelays,
   setZapForward,
   setSensitive,
   reset,
@@ -34,6 +35,7 @@ import { EventBuilder } from "System";
 import { Menu, MenuItem } from "@szhsin/react-menu";
 import { LoginStore } from "Login";
 import { getCurrentSubscription } from "Subscription";
+import useLogin from "Hooks/useLogin";
 
 interface NotePreviewProps {
   note: TaggedRawEvent;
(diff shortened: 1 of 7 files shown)


packages/app/src/Element/NoteCreator.tsx
```diff
@@ -232,6 +238,38 @@ export function NoteCreator() {
...
   }
+  
+  function renderRelayCustomisation() {
+    return (
+      <div>
+        {Object.keys(relays.item || {}).filter((el) => relays.item[el].write).map((r,i,a) => 
+          <div className="card flex">
+            <div className="flex f-col f-grow">
+              <div>
+                <FormattedMessage defaultMessage="{relay}" values={{ relay:r }} />
```
v0l, line 249: Don't think there is any reason to wrap this in `FormattedMessage`, there is nothing to localize

packages/app/src/Element/NoteCreator.tsx
```diff
@@ -330,6 +368,13 @@ export function NoteCreator() {
               <button className="secondary" onClick={loadPreview}>
                 <FormattedMessage defaultMessage="Toggle Preview" />
               </button>
+              <h4>
+                <FormattedMessage defaultMessage="Custom Relays" />
+              </h4>
+              <p>
+                <FormattedMessage defaultMessage="Send note to a subset of my write relays" />
```
v0l, line 375: change `my` with `your`

packages/app/src/Element/NoteCreator.tsx
DanConwayDev: agreed

packages/app/src/Element/NoteCreator.tsx
DanConwayDev: changed

v0l:

PR #537, opened 2023-04-20
Fix zine links without trailing slash, ignore top and new links

blastshielddown:


packages/app/src/Const.ts
```diff
@@ -178,4 +178,4 @@ export const MagnetRegex = /(magnet:[\S]+)/i;
  * Wavlake embed regex
  */
 export const WavlakeRegex =
-  /(?!zine\.wavlake\.com)(?:player\.|www\.)?wavlake\.com\/(?:(?:track|album)\/[a-f0-9]{8}(?:-[a-f0-9]{4}){3}-[a-f0-9]{12}|[a-z-]+)$/i;
+  /https?:\/\/(?:player\.|www\.)?wavlake\.com\/(?!top|new)(?:(?:track|album)\/[a-f0-9]{8}(?:-[a-f0-9]{4}){3}-[a-f0-9]{12}|[a-z-]+)/i;
```
SamSamskies, line 181: https://www.wavlake.com/artists renders a broken embed. Could you just make sure the link contains `track` or `album` instead?

```suggestion
  /https?:\/\/(player\.|www\.)?wavlake\.com\/(track|album)\/([a-f0-9]{8}(-[a-f0-9]{4}){3}-[a-f0-9]{12}|[a-z-]+)/i;
```

SamSamskies (changes requested):
you committed a package-lock.json by accident.

packages/app/src/Const.ts
blastshielddown: Problem is our individual artists pages need to be matched too and the patterns for those are like `wavlake.com/artist-name`. We could reject match on `artists` path as well to be safe

SamSamskies: ah damn. that's means you'll have to update the regex every time you add a new path.

blastshielddown: Exactly. It's not ideal, but we don't have new paths pop-up very often. What do you think of this latest version? Covers all the existing cases (and a few that I don't see why someone would share but just in case).

SamSamskies (approved):
🚀

SamSamskies:
I'm not sure why the linting is failing. I don't think it has anything to do with what you did, but could you try rebasing main and then running `yarn workspace @snort/app format`?

blastshielddown:
Strange. Ran prettier and linting comes out OK locally. I also tried splitting the regex into 2 lines to see if that would help -- no dice. Also tried a completely new branch from main and PR'd from that one -- same error. Not sure what the issue is here

SamSamskies:
@blastshielddown you have to rebase `main` first and then run prettier. the lint error is not about the changes you made. not sure why it's doing that.

SamSamskies:
you need to make sure to fetch the updates from the `main` branch in the snort repo.

blastshielddown:
My force push triggered closing this, reopening now (hoping i did the rebase right)

PR #542, opened 2023-04-22
fix text overflow on follow button

vivganes:
In some languages and resolutions, the text 'Follow' for hashtags is overflowing.

image

This is due to the fixed width of the button.

This PR fixes it.

Screenshot after fix:

image

### DIFF
--- a/packages/app/src/index.css
+++ b/packages/app/src/index.css
@@ -601,6 +601,10 @@ button.tall {
   justify-content: space-between;
 }
 
+.action-heading button {
+  min-width: 98px;
+}
+
 .flex-end {
   display: flex;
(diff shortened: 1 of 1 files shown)


packages/app/src/index.css
```diff
@@ -601,10 +601,6 @@ button.tall {
   justify-content: space-between;
 }
 
-.action-heading button {
-  width: 98px;
```
v0l, line 605: Maybe use `min-width` instead, the fixed width is important

v0l (changes requested):

packages/app/src/index.css
vivganes: done

PR #547, opened 2023-04-29
fix url regex

hellodword:
For example:
`https://en.wikipedia.org/wiki/Git_(slang)`
[`https://en.wikipedia.org/wiki/Git_(slang)`](https://en.wikipedia.org/wiki/Git_(slang))

snort will match `https://en.wikipedia.org/wiki/Git_(slang`


v0l:
Added, thanks

PR #549, opened 2023-05-08
Fix relay list

SnowCait:
`r` tag without second parameter (which means `read: true, write: true`) is added to kind 10002 when `read: false, write: false`.


v0l:
Fixed in 9f8095b0df848f92ce9e828e2c52cc8d9a71cd03

PR #569, opened 2023-09-27
Rebranding features

mmalmi:
- config/*.json files for building with another appName or other settings
- `NODE_ENV_CONFIG=iris yarn build` for iris-branded build
- FormattedMessage component replaces Snort with config.appName in translations
- Submit PIN prompt by pressing enter
- {config.nip05domain}/{username} route support
- If iris.myKey is in localStorage, convert it to a Snort key with PIN 1234 (I might want to disable PIN protection for Iris build later)


packages/app/src/Login/MultiAccountStore.ts
```diff
@@ -217,6 +218,23 @@ export class MultiAccountStore extends ExternalStore<LoginSession> {
     return { ...s };
   }
 
+  async #loadIrisKeyIfExists() {
+    try {
+      const irisKeyJSON = window.localStorage.getItem('iris.myKey');
+      if (irisKeyJSON) {
+        const irisKeyObj = JSON.parse(irisKeyJSON);
+        if (irisKeyObj.priv) {
+          const privateKey = await PinEncrypted.create(irisKeyObj.priv, '1234');
```
v0l, line 227: Some people dislike the pin feature, im planning to make it optional, if you want to store an unencrypted in you can make another class which just has `get value()` then it should work the same, and i can also implement this later in the UX so people can stop complaining about it

PR #571, opened 2023-10-16
reset relay publish status screen if it is closed for 5 seconds

fiatjaf:
Currently if you click outside the modal it will disappear, but once you try to write a new note it will show up again with the relay statuses of your previous publish.


v0l:
Fixed

PR #572, opened 2023-10-16
don't close publish modal automatically

fiatjaf:
currently if one of the target relays fail you get a nice modal with a list of all successes and failures, but if all relays succeed your screen blinks with an unreadable modal and you don't know to where the note was published.

with this even if all relays responded with an OK the modal will still remain open so the user can look at the list of relays the note was published to.


v0l:
There was supposed to be a delay actually but i guess i forgot to add it

PR #574, opened 2023-11-11
Trim Zap comment

benthecarman:
We had a bug when trying to parse the json of a comment. It had a \n at the end of it and a browser's json parser couldn't handle it. This should fix


v0l:
merged in 9b5721faa401284a0d0b7134d51f146c0a620958

PR #575, opened 2023-11-21
Enhancements/593-validation

trycatchkamal:
Kieran/snort#593

✅ NIP-05 test
✅Lightning Address test
✅Limit username length(100)
✅Limit about length(1000)


packages/app/src/Pages/settings/Profile.tsx
```diff
@@ -112,6 +125,103 @@ export default function ProfileSettings(props: ProfileSettingsProps) {
...
+        setAboutValid(false)
+        setInvalidAboutMessage(formatMessage(messages.AboutLengthError))
+      }else{
+        setAboutValid(true)
+        setInvalidAboutMessage("")
+      }
+    }
+  }
+
+  async function nip05NostrAddressVerification(nip05Domain: string | undefined, nip05Name: string| undefined) {
```
v0l, line 170: I think we have a function for this already, like `fetchNip5Pubkey` or something like that in `@snort/shared`

packages/app/src/Pages/settings/Profile.tsx
```diff
@@ -112,6 +125,103 @@ export default function ProfileSettings(props: ProfileSettingsProps) {
...
+          setNip05AddressValid(false)
+          setInvalidNip05AddressMessage(formatMessage(messages.ErrorValidatingNip05Address))
+      }
+    }catch(e){
+      setNip05AddressValid(false)
+      setInvalidNip05AddressMessage(formatMessage(messages.InvalidNip05Address))
+    }
+  }
+
+  async function onLud16Change(address:string){
```
v0l, line 200: This would be better if it was triggered with `useEffect` and use `return debounce(() => fn)` to add a delay, this is what we do for search etc, you should be able to see this pattern used already

packages/app/src/Pages/settings/Profile.tsx
```diff
@@ -12,8 +12,11 @@ import { UserCache } from "@/Cache";
 import useLogin from "@/Hooks/useLogin";
 import Icon from "@/Icons/Icon";
 import Avatar from "@/Element/User/Avatar";
-import { FormattedMessage } from "react-intl";
+import { FormattedMessage, useIntl } from "react-intl";
 import { ErrorOrOffline } from "@/Element/ErrorOrOffline";
+import { LNURL, LNURLError } from '../../../../shared/src/lnurl';
```
v0l, line 17: this import should be `@snort/shared`

v0l (changes requested):
Sorry for the slow response here, completely missed snort PR's, DM me on nostr if im too slow

trycatchkamal:
> Sorry for the slow response here, completely missed snort PR's, DM me on nostr if im too slow

@v0l yeah thought of dropping a message, but I wasnt sure if you were looking at the DMs. I had already tried to get in touch with you regarding some queries on snort local setup but then ended up getting sorted myself. 

Will do drop a message from now on! Thanks

packages/app/src/Pages/settings/Profile.tsx
```diff
@@ -112,6 +146,67 @@ export default function ProfileSettings(props: ProfileSettingsProps) {
...
+        setAboutValid(false)
+        setInvalidAboutMessage(formatMessage(messages.AboutLengthError))
+      }else{
+        setAboutValid(true)
+        setInvalidAboutMessage("")
+      }
+    }
+  }
+
+  async function nip05NostrAddressVerification(nip05Domain: string | undefined, nip05Name: string| undefined) {
```
v0l, line 191: This should also use `useEffect` with the `debounce` otherwise its going to make a request for every char that you enter or delete, you will also have a problem when 1 request overwrites the result if its slower/faster.

v0l (changes requested):

PR #576, opened 2023-11-30
Show border when there is no selected DM

trycatchkamal:
Fixes https://git.v0l.io/Kieran/snort/issues/710

* Shows red border as suggested when there is no selected DM and is visible only on  large screens

### DIFF
--- a/packages/app/src/index.css
+++ b/packages/app/src/index.css
@@ -932,6 +932,6 @@ svg.zap-solid {
 @media (min-width: 768px) {
   .rt-border{
     border-right-width: 1px;
-    border-right-color: red;
+    border-right-color: var(--border-color);
   }
 }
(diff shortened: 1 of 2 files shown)


packages/app/src/index.css
```diff
@@ -928,3 +928,10 @@ svg.zap-solid {
 .hide-scrollbar::-webkit-scrollbar {
   display: none; /* Safari and Chrome */
 }
+
+@media (min-width: 768px) {
+  .rt-border{
+    border-right-width: 1px;
+    border-right-color: red;
```
v0l, line 935: I dont think it was supposed to be red, thats just for emphasis it should be `var(--border-color)` probably

v0l (changes requested):

v0l (approved):

PR #578, opened 2024-01-01
change settings message when choosing a username

eesazahed:
It says the username must be between 1 and 32 characters. It should be 8 and 15. Iris will only accept it with that character length.

Screen Shot 2024-01-01 at 4 53 57 PM


mmalmi:
Thanks for noticing, fixed!

PR #592, opened 2024-11-13
[worker-relay] Decouple worker functions from worker

dskvr:
This decouples the worker logic, namely the batching and message handling (and also types) from the worker.

**Why?**

_tldr;_

Flexability. Enables usage of everything in this library except the included `WebWorker`, with minimal changes to the package and no difference in usage.

....

It makes this package more flexible. In my case, I when I add an event, I don't do so from the main thread. I add the event from another web worker over a `MessageChannel`. While _maybe_ this package should include message channel functionality in the `init` message, I feel like that might be out of scope, and it does open a can of worms. 

By decoupling the worker logic from the worker itself, I am able to import just the worker logic I need, implement it according to my needs, while retaining the usefulness of `WorkerRelayInterface` for the main thread. 

**Notable Changes**
- All worker state was moved to a `state` object (typed) so that changes to state could be made by reference.
- `wasm:copy` script in package.json that copies `wasm` to dist/sqlite
- exported all relay actions so that the relay could receive commands without mocking `MessageEvent`'s first. (note: less readable, less maintainable, less clean than it was previously, but way more versatile). Added around ~100 bytes to package.

**Notes**
- Will cleanup history and squash  if you decide you want this, accidentally included changes from my other PR 🤦 
- **MIGHT** cause issues in some situations when bundling, but at a glance, the `worker` should bundle correctly in for `es/iife` and `inline`

**Potential Improvements**
Expose via exports in package instead of dist traversal on import.

**Usage Changes**
None. (see disclaimer)

**Disclaimer**
100% untested, hence the draft. I've made these same changes in the fork I'm maintaining so will likely patch and confirm functionality later.

### DIFF
--- a/packages/worker-relay/example/basic.ts
+++ b/packages/worker-relay/example/basic.ts
@@ -1,7 +1,7 @@
-import { WorkerRelayInterface } from "@snort/worker-relay";
+import { WorkerRelayInterface } from "@nostrwatch/worker-relay";
 
 // when using Vite import the worker script directly (for production)
-import WorkerVite from "@snort/worker-relay/src/worker?worker";
+import WorkerVite from "@nostrwatch/worker-relay/src/worker?worker";
 
 // in dev mode import esm module, i have no idea why it has to work like this
 const workerScript = import.meta.env.DEV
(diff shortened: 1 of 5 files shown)


dskvr:
In order for `MessageChannels` to work requires a different approach, namely because it still requires an `init` event to set the `MessageChannel`'s `port`. 

My fork has diverged in this area via some ugly patches, when i find a solution that doesn't mess with existing functionality or make this unmaintainable, will open another PR.

PR #595, opened 2024-11-18
[worker-relay] Add PRE and fix RE support.

dskvr:
Was treating everything between 10000 and 40000 as a Replacable Event, this effectively broke PRE's because any given d tag couldn't exist more than once in the worker relay, regardless of which pubkey it was from.

There wasn't any noticeable performance loss on a database with 20k events (other than the whole 20k events thing)

Todo:
- [ ] Don't store ephemeral events?

### DIFF
--- a/packages/worker-relay/src/sqlite/sqlite-relay.ts
+++ b/packages/worker-relay/src/sqlite/sqlite-relay.ts
@@ -34,7 +34,7 @@ export class SqliteRelay extends EventEmitter<RelayHandlerEvents> implements Rel
     await this.#open(path);
     if (this.db) {
       await migrate(this);
-      // dont await to avoid timeout
+      // don't await to avoid timeout
       runFixers(this);
     }
   }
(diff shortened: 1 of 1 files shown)


dskvr:
@v0l this one works well and is ready, but it also might introduce breaking changes into snort.

packages/worker-relay/src/sqlite/sqlite-relay.ts
```diff
@@ -132,69 +132,82 @@ export class SqliteRelay extends EventEmitter<RelayHandlerEvents> implements Rel
...
+      const oldEvents = db.selectValues(
+        `SELECT id FROM events WHERE kind = ? AND pubkey = ?`,
+        [ev.kind, ev.pubkey]
+      ) as Array<string>;
+
       if (oldEvents.includes(ev.id)) {
-        // we already have this event, return
+        // Already have this event
         this.#seenInserts.add(ev.id);
-        if (oldEvents.length > 1) {
```
dskvr, line 145: @v0l This was causing problems for me, but you likely had it here for a reason...

packages/worker-relay/src/sqlite/sqlite-relay.ts
```diff
@@ -132,69 +132,82 @@ export class SqliteRelay extends EventEmitter<RelayHandlerEvents> implements Rel
...
-    if (legacyReplacable.includes(ev.kind) || (ev.kind >= 10_000 && ev.kind < 20_000)) {
-      const oldEvents = db.selectValues("select id from events where kind = ? and pubkey = ? and created <= ?", [
-        ev.kind,
-        ev.pubkey,
-        ev.created_at,
-      ]) as Array<string>;
+    const legacyReplaceableKinds = [0, 3, 41];
+
+    // Handle legacy and standard replaceable events (kinds 0, 3, 41, 10000-19999)
+    if (legacyReplaceableKinds.includes(ev.kind) || (ev.kind >= 10_000 && ev.kind < 20_000)) {
```
dskvr, line 138: this looks wrong to me, need to go back and look at this.

dskvr:
I need to add the `created_at` safeguards back into into RE and add to PRE queries, because for some dumb reason I removed them.

dskvr:
Ok, I was mistaken. I moved the safeguards to memory, as it made the implementation more linear. Please double check this change during review.

packages/worker-relay/src/sqlite/sqlite-relay.ts
```diff
@@ -132,69 +132,82 @@ export class SqliteRelay extends EventEmitter<RelayHandlerEvents> implements Rel
...
+      const oldEvents = db.selectValues(
+        `SELECT id FROM events WHERE kind = ? AND pubkey = ?`,
```
v0l, line 140: Missing the `created` check, if we insert RE's out of order or we get older version last we will be inserting that one over the newer event already stored

packages/worker-relay/src/sqlite/sqlite-relay.ts
v0l: There was a bug, there was multiple copies of the RE's and they didn't get deleted because there was no newer version to replace them.

The `> 1` check was there to delete the older versions, maybe its not needed now because most should have this old data cleaned up by now

packages/worker-relay/src/sqlite/sqlite-relay.ts
```diff
@@ -132,69 +132,82 @@ export class SqliteRelay extends EventEmitter<RelayHandlerEvents> implements Rel
...
+      const dTag = ev.tags.find(a => a[0] === "d")?.[1] ?? "";
+
       const oldEvents = db.selectValues(
-        "select id from events where id in (select id from events, tags where events.id = tags.event_id and tags.key = ? and tags.value = ?)",
-        ["d", dTag],
+        `SELECT e.id
+         FROM events e
+         JOIN tags t ON e.id = t.event_id
+         WHERE e.kind = ? AND e.pubkey = ? AND t.key = ? AND t.value = ?`,
+        [ev.kind, ev.pubkey, "d", dTag]
```
v0l, line 163: Wow, not sure how i missed this being completely wrong, maybe i only worked with events which had unique `d` tags

dskvr: Happens

packages/worker-relay/src/sqlite/sqlite-relay.ts
dskvr: @v0l added for both PRE and RE old events check.

v0l:
Can you rebase this

v0l (approved):

dskvr:
rebased and squashed the commits.

PR #596, opened 2024-12-06
feat: Add config for soloco.nl

kroese:
Add config for soloco.nl


v0l (approved):

v0l:
ah it might be force pushed from git.v0l.io

